PR #1, opened 2021-04-18
Basic header and footer

vincerubinetti:


3b1b:
Looking good!  Small notes:

- I'd make the name 3Blue1Brown appear in all white, rather than coloring the numbers.
- Once there's content, I'd make the whole banner with the logo smaller, maybe even 1/3 of the height it now is, clearing the way for the landing page to be more dominated by content. 
- Maybe the logo for "lessons" should be the letter π instead of sqrt(x)

3b1b:
As to subtle animations in line with the kind of things you've done before, I'm open to seeing whatever you have in mind.  One possibility could be that the grid appears in a subtle 3d perspective which rotates to make the whole thing flat as one scroll down.

I also think the best opportunity for this kind of animation would be at the tops of various articles.

vincerubinetti:
> maybe even 1/3 of the height it now is

FWIW, I was only going to have it that large on the home page. On every other page, the logo  was going to collapse into the center, sort of like [this](https://mygeneset.info/) then [this](https://mygeneset.info/about). Not sure if that seems better, but I think it's okay to have your logo a little big at least on the home page to establish branding and let people know they're in the right place.

> π instead of sqrt(x)

I like to use this library called [Font Awesome](https://fontawesome.com/icons?d=gallery&p=2&q=math) for icons, since they're professionally designed and consistent with each other. So far that's what I'm using here. Unfortunately π is behind a paywall. Is there any other icon there you'd prefer for lessons? If not I can just get π from somewhere else.

👍 on everything else

3b1b:
> I think it's okay to have your logo a little big at least on the home page to establish branding and let people know they're in the right place.

I see what you mean for the home page.  It might be one of those things that are easiest to judge once there's more content there.

> Is there any other icon there you'd prefer for lessons?

Well, a lot of the usual icons for "math" elicit feelings of school math, whereas the brand I'd like to go for would be more around beautiful math.  Hard to capture in a single icon, maybe for want of creativity π is the only one I can think of.  Would any of [these](https://thenounproject.com/search/?q=pi) fit the style?

vincerubinetti:
> elicit feelings of school math

Very good point. Noun Project will have something usable I'm sure. I'll find one that's similar enough that no one will notice except me 😄

3b1b:
Also, I'm wondering how absurd it would be to have some easter egg where upon clicking the logo (or something even more hidden) it does [this](https://www.dropbox.com/s/9kx2qm6x7xl49yo/StreamIntro.mp4?dl=0), as generated by the "StreamIntro" animation [here](https://github.com/3b1b/videos/blob/master/_2020/ldm.py).

I think the answer might be "pretty absurd, definitely not worth doing".

vincerubinetti:
Your wish is my command. I'm glad you're looking to do those things because I like doing them. Doing exactly what's in that animation might take more time and code than it's worth, but something similar but simplified I could do in a couple minutes now that we have the clean SVG.

I was thinking of having other fun easter eggs like something special on 3/14, though further down the road once the critical stuff is done.

3b1b:
The beauty of it is that if the spikes are each treated as independent triangles, it's actually a very small amount of code, i.e. each one is just set to rotate at a rate proportional to the initial angle of its tip.

But of course, if it's anything other than very straightforward it's silly enough to not be worth doing.

vincerubinetti:
Put an item for this here:
https://github.com/vincerubinetti/3Blue1Brown.com/projects/1?add_cards_query=is%3Aopen

We can come back to it. Not immediately obvious to me at this moment how I'd do this in CSS cleanly (it'd be easy in Javascript but you generally don't want to use that for DOM animations for various reasons).

vincerubinetti:
Going to merge this for now because I want to see -- for a different project I'm working on -- if the GitHub Pages build fails. Not sure if the header animation is that great but we can come back to it.

PR #3, opened 2021-04-24
homepage layout

vincerubinetti:
- basic homepage layout
- add basic card component
- add basic figure component
- split _includes into folders
- add slideshow component
- add subscribe to header nav
- add material design color palette
- add spotlight, topics, follow, and support sections to homepage
- pull in store images from DFTBA and make transparent
- import pi creature SVGs from Ben Eater site and clean up
- add fade in plugin for fading in arbitrary items when they come into view


vincerubinetti:
@3b1b this is ready for you to take a look at. Some thoughts:

1. I'm not 100% happy with the header grid animation. Will continue brainstorming. There are many possible ideas (hilbert curve? fourier circles?), but it should be something non-distracting, and also non-specific enough that it can sort of apply to your whole brand and website
2. Maybe we can liven up the "spotlight" section a bit more with the pi creatures?
3. Perhaps we can combine the header and spotlight sections into one, and make it more artsy, if you're into it. Something like... we start off with a fullscreen black page, logo and text fade in, then logo moves to top and video fades in, then when user scrolls/interacts, the nav bar appears, the logo condenses to the top, and the user can see the rest of the page. All of this would be quick and responsive as to not frustrate users who just want to scroll down to content right away.
4. Can you send me a high-res and high-framerate mp4 of the pi fourier for the bottom. I can compress it down as necessary.

3b1b:
General feedback, don't worry about staying too close to the layout or design of the current website.  Feel free to go back to first principles.

> I'm not 100% happy with the header grid animation....

Actually, I quite like it.  It fits the goal of being generic but still on-brand.  I still think it should be smaller though, even half the height.  I would say we should first prioritize having a good scaffolding for how we'll navigate into all the content articles and explore what's there.

> Maybe we can liven up the "spotlight" section a bit more with the pi creatures?

Looks good for now, but if we're going to embed the video there let's make it large enough to be comfortably viewable, perhaps even as tall at the spotlight section itself.  It's unclear if that section even needs a title.  Can we set some of the [embed arguments](https://developers.google.com/youtube/player_parameters) on that video so that the controls and title don't show up; i.e. the view frame should have nothing but the video content itself.

> Perhaps we can combine the header and spotlight sections into one, and make it more artsy, if you're into it. Something like... we start off with a fullscreen black page, logo and text fade in, then logo moves to top and video fades in, then when user scrolls/interacts, the nav bar appears, the logo condenses to the top, and the user can see the rest of the page. All of this would be quick and responsive as to not frustrate users who just want to scroll down to content right away.

I think just landing directly on the content is preferable, which comes from the same part of my brain that wants a smaller header.

> Can you send me a high-res and high-framerate mp4 of the pi Fourier for the bottom. I can compress it down as necessary. Later, if you want to do something fun with this, tell me the periods and radii of the circles and I can have javascript draw it with SVG and react to mouse/scroll movements or something (I could do this relatively quickly).

Sure, I'll email you all that.

3b1b:
Also, some other notes:

- I'm happy to eliminate the support and store sections from the home page for now, letting those live more subtly in the form of header links.  Alternatively, we could have a section right below "follow" in the same style that says something like "Want to help the project?", with three icons, "share", "contribute", "shop".  Again, all would be in the service of making the content front and center with all ancillary things sitting more at the periphery.
- Can we remove the icons from the header links?

vincerubinetti:
Sounds good, take a look at the updated preview. 

Two items I have questions about:

1. The only embed options that seem relevant are `autoplay`, `controls`, `fs`, and `modestbranding`. Turning controls off and modestbranding doesn't seem to have the desired effect. The video title, share, watch later, playlist, and youtube buttons are all still there. `controls` only seems to turn off the scroll bar. I don't think you want to turn this off because it'll frustrate users wanted to rewind and etc. Also regarding making it bigger, I did, but I assumed we would keep the fullscreen button to allow people to watch it more... fully.
2. Do you actually want to have the fourier pi at the bottom then? Or should we save it for the actual Fourier series/topic. It seems like you're going for a quite minimal approach.

vincerubinetti:
Regular embed:
image

`&controls=0&fs=0&modestbranding=1`:
image

3b1b:
Yeah, see there's only so much one can do with the embedded player.  The main gripe I have (which is with YouTube and not you) is the title and everything on the top.  I guess people are used to it now, knowing that full screen or clicking over to YouTube are the ways to watch it without clutter.

As to the pi Fourier, let's just wait until the Fourier Series article exists.  Maybe we could have a "featured content" shelf on the home page, and some nice animation with the javascript you described could be the thumbnail for that particular topic, but it's a bridge to cross when we get there.

vincerubinetti:
Yeah I dislike YouTube's embed as well. I've looked for alternatives to it, but I don't think they're open-minded enough to allow something like that. Best we could do is have a cleaner thumbnail with just a play button that replaces itself with the real YouTube embed on click. For now I've reverted to having the scroll bar.

👍 re: fourier. Will keep it in the backlog of fun things to do when we're done with everything else.

PR #29, opened 2021-05-11
Topic reorg

3b1b:
A general reorganization of all the article stubs
- URLs have been abbreviated
- Different categorization and ordering
- Added topics-organization.yaml to match the organization.  Currently this is not used, but it will serve as a more conveniently editable source of ground truth for the full article organization.


vincerubinetti:
I think the `?` in a filename may have caused the last build failure?

Also you may want to delete all those useless Netlify notifications and emails and enable just:

- `GitHub commit status → Deploy Preview succeeded`
- `GitHub commit status → Deploy Preview failed`
- `GitHub pull request comment → Deploy Preview succeeded`
- `GitHub pull request comment → Deploy Preview failed` 

And perhaps rename the site to `3blue1brown-website` off the randomly generated name `focused-rosalind-cc7334`.

SnFyCh0Z28

vincerubinetti:
@3b1b I recommend removing the text from the topic thumbnails if possible: 1. text will then be vector and scale nicely 2. image and text separated, making both more readable and image less busy 3. design more consistent; text always in same place 4. easier to change text position and design later if we want 5. slightly easier for me to make accessible for screen readers

3b1b:
Sounds good, I'll remove the text.  Is your thought then to display the same white CMU serif font text on top of the icons there so they look the same, or were you also suggesting having the title text sit outside the image?

vincerubinetti:
Personally I think outside the image, and not on top, looks nicer and cleaner. But we can always change that later.

Also I made a note in another issue: Make sure to overshoot the resolution of these images. Go higher than you think you need to.

vincerubinetti:
@3b1b If you're done with this, looks good and ready to merge. I intend to do some work tonight on the site so it'd be good to have your new lesson order file there to work with.

PR #31, opened 2021-05-12
Added OSX installation steps for local site to README.md

KurtBruns:
These are the steps that Grant, Josh and myself followed to get the local Jekyll site running on our machines.

@vincerubinetti can you make sure we didn't do anything weird? so we don't run into trouble later. There are two commands in particular to look at.

When installing bundler we used the `user-install` flag rather than using `sudo`.

```sh
gem install --user-install bundler
```

After encountering an error when running `bundle install` we set the configuration to be local.

```sh
bundle config set --local path 'vendor/bundle'
```

I linked to resources I found online for troubleshooting errors. Also, for reference, I'm on a MacBook Pro running Catalina and the two other machines I think are the new M1 Macs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,3 +20,77 @@ To make more structural changes to the site, you may want to build it locally:
 
 The site should automatically open in your default browser.
 Any changes you make should automatically rebuild the site and refresh the page, except for changes to `_config.yml` which require re-running the start command.
+
+### OSX Installation Notes
+
+Install ruby using [homebrew](https://brew.sh/).
+
+```sh
+brew install ruby
+```
+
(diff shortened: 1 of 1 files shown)


vincerubinetti:
This shouldn’t affect the site in any way, feel free to merge. I’m curious why this is necessary though because the installation is simpler on windows and Linux.

KurtBruns:
@vincerubinetti I don't know why we ran into so many errors. I think some were related to the new M1 machines and others to the Mac OS rootless security feature.

Anyways, it was difficult enough to feel worthwhile to document the setup process.

vincerubinetti:
Sorry about all that difficulty, it should've been much easier. I'm hoping it's due to the fact that you're on new hardware, and that Jekyll/Ruby will fix these issues soon.

PR #33, opened 2021-05-12
Create applet functionality

PullJosh:
Experimenting with embedding interactive applets into articles. This is quick and dirty, and might be a totally ridiculous way of doing things. I have no idea what the Jekyll conventions are. 🤷‍♂️

@vincerubinetti, feel free to reject and/or totally restructure this if I'm doing it wrong.

### DIFF
--- a/_includes/components/applet.html
+++ b/_includes/components/applet.html
@@ -1,0 +1,5 @@
+<div class="applet">
+  <div class="applet_aspectRatio">
+    <iframe class="applet_iframe" src="{{ include.path | relative_url }}"></iframe>
+  </div>
+</div>
--- a/_posts/2016/2016-11-04-who-cares-about-topology-inscribed-rectangle-problem.md
+++ b/_posts/2016/2016-11-04-who-cares-about-topology-inscribed-rectangle-problem.md
@@ -9,3 +9,7 @@ source: _2016/wcat.py
 credits:
 - Lesson by Grant Sanderson
 ---
+
+Hello, world!
+
+{% include components/applet.html path="applets/2016/2016-11-04-who-cares-about-topology-inscribed-rectangle-problem/surface-sketcher.html" %}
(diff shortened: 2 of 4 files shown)


PullJosh:
Totally love the live previews. Here's the page with the applet if you just want to check it out: https://deploy-preview-33--3blue1brown.netlify.app/lessons/who-cares-about-topology-inscribed-rectangle-problem.html

vincerubinetti:
This looks nice. This is basically how I would do it for iframes, yeah. I might tweak the CSS a little.

But I'd say let's hold off on this, at least until we sit down and have a more formal discussion about it, because there's a couple ways to do it that might affect things down the road. 

For example iframes are completely sandboxed and offer no way to style or control the contents of the iframe from the actual page it's displayed on, which might be wanted later. Another approach could be having each applet be a react or similar app that exports compiled JS/CSS/etc and just get mounted to a container div in a jekyll component. This could allow things like more flexible layouts, styling all applets at the same time and in the same way from the website, extra controls added outside the applet, etc.

Much to think about, let's put a pin in it until the more critical stuff is done. Unless you plan to have these visualizations in like every lesson in which case we can talk about it sooner.

3b1b:
This is wonderful.  I think sprinkling in a few things like this through to the site will be worthwhile, as long as none of them suck us in too much.  For now, to Vince's point, let's wait until the rest of the site's features are in place.

A few notes on the interactive itself
- I would turn off the ambient rotation.
- If there are any options for multisample antialiasing or something to that effect, it could help it look nicer.
- I wonder if it's nicer for the canvas you draw on to be separate from the graph (above?) with a background color matching the xy-plane in the 3d plot.

PullJosh:
Let's put a pin in this for now, then. It definitely isn't urgent.

To @vincerubinetti's point: Writing these as React components would be awesome! I have no idea how integrating React components into Jekyll would work, but it sounds great.

@3b1b Those are great ideas! The background color to match the plane is really smart.

PR #35, opened 2021-05-15
Lesson gallery

### LINKED ISSUE #14: Lesson "featured", "by date", and "by topic" tabs
Splitting this out from #4 into it's own separate issue. Maybe have a button to toggle between showing the topic cards, and just a chronological listing of lesson cards. Wouldn't actually be too hard to do.

### LINKED ISSUE #16: Spotlight fixes
Pi creatures get covered up by video on small screens. Just remove.

Also per meeting on 5/9 we are getting rid of spotlight all together? Are we sure? Then there will be nothing directly watchable on the homepage.

### LINKED ISSUE #17: Collapse header nav menu

### LINKED ISSUE #18: Figure component tweaks
Don't autoplay video until switch to that tab. Pause video when switching back to image tab.

### LINKED ISSUE #20: Question component tweaks
Don't show correct answer when selecting wrong one. Show reset button after submitting.

### LINKED ISSUE #28: Emotive Pi Characters
Today we were talking about the value of the pi characters for expressing emotion.

It would be really great to have one or more components that enable us to use the pi characters to communicate emotion and whimsy. In particular, we discussed two ways the pi characters might be used:
1. To personify footnotes that contain little side points. Not all footnotes would have pi characters, but some could. It's worth considering this while implementing any footnote funcitonality.
2. Used inline with the regular body text to emphasize an emotional point. For example:
image
It would be cool for a pi character to make this exclamation.

Any component for these characters should definitely make it possible to specify the emotion of the pi character (as well as the associated text).

### LINKED ISSUE #30: Spruce up 404 page
Add sad pi with heart broken. Maybe something else fun.

Include site-wide search when we have that ready per #24

### LINKED ISSUE #32: Discussion of image scaling and page width
If we change the content width on the site to be a number with more factors than `1000`, such as `960` or `1080`, then it will make designing grid layouts and custom illustrations a little easier and IMO more pleasing. Namely, `1000` is missing the factor of `3`.

**Note:** This is pretty nit-picky since 1000px is perfectly fine. 

To illustrate why this might be nice, consider that `960` can be evenly divided by common numbers and the results also have a lot of factors too.

```
960/2 = 480
960/3 = 320
960/4 = 240
960/5 = 192
960/6 = 160
```

For completeness, shown below are the prime factorizations of the three numbers 960, 1000 and 1080.

```
960=2*2*2*2*2*2*3*5
```

```
1000=2*2*2*5*5*5
```

```
1080=2*2*2*3*3*3*5
```

### LINKED ISSUE #36: Top of the lesson page layout
I'm just recording a few small notes here so I don't forget.

- Arguably the video embed should go above the title, dates, byline, etc. in the top of the lesson.
    - That way, when a person clicks on a lesson, what dominates the screen is an actual lesson rather than descriptors.
    - Having the title below the video also matches how its shows up on YouTube.  Not that they need to match, but it's a positive signal for the video-at-the-very-top design.
    - The title, dates, byline, etc. then serve as a kind of separator between video content and written content.
- Would it be crazy to have the video fill the full width of the page, analogous to the YouTube player's "theater mode"?
- Instead of displaying the group of the video below the byline, save the screen real estate and just don't display it at all, but maybe show it if someone clicks to open the lesson list (probably shown at the top of the list).
- Given that realistically it will take a while for most content to be bloggified, when there is no content in the markdown body (or even just if it's less than some threshold), we should not show the previous/next lesson bar at the bottom of the page due to the redundancy with the one at the top.
- Don't show the updated date when there's only a video and no written content.

Screen Shot 2021-05-14 at 11 00 11 PM

vincerubinetti:
closes #36 
closes #32 
closes #30 
closes #28 
closes #20 
closes #18 
closes #17 
closes #16 
closes #14 

- touch up 404 page
- update readme
- move testbed post to future so it shows up in local test builds but not live site
- add featured lesson list
- make new topics list from Grant's prev reorg
- rename some css classes to be more consistent
- autoplay figure video when it comes into view, either by scrolling or clicking to video tab
- add basic pi creature component
- refactor question component to have "try again" button and not show correct answer on wrong answer
- de-generalize clickable component a bit, remove from header and footer
- make header expandable/collapsible on mobile
- refactor lesson card component. add date. reference by id (eg `linear-algebra-1`)
- make lesson gallery sortable by feature, topic, date
- rename `expanded` to simpler `open`
- delate date, topic, and order fields from lesson metadata
- several css tweaks, simplifications, refactors
- fix spotlight pis, and make use of new pi component
- move topic detail to lesson nav bar list button
- make lesson layout changes. remove elements when unnecessary
- rename table of contents to lesson nav


3b1b:
Looks good to me.  What's the best way to see the testbed-lesson now, I ended up just copying the markdown into a different lesson.

I'll be out of town for the weekend, but if there's a question I can answer regarding pi creatures feel free to let me know.

vincerubinetti:
Ah yes, you'll have to add the `--future` flag to the `jekyll build` comment in Vercel to get that to show up (in the "by date" tabs).

Also if you're running it locally, just run `start.sh` which includes all the appropriate flags.

vincerubinetti:
Commits since my last comment:

- Implement pi creature component options. Can put it to the side of the main page column, or inline with the text as its own paragraph, or "auto" to to the former when the screen is wide and the latter when the screen is mobile. We'll still going to want to add speech/thought bubbles, but let's do that in another pr.
- Implement your change requests in #36 
- other misc changes

3b1b:
Looks good to me, thanks!

PR #47, opened 2021-05-25
Workflow

KurtBruns:
This isn't ready to merge, but I'm opening up this pull request to start testing Vercel. 

The current branch variable can be used as follows for building the `git lfs` URL in the **development** environment.

```
.Site.Branch
```

The bucket URL can be used for prefixing the Linode Bucket in the **production** environment, defaults to an empty string for **local** and **staging**.

```
.Site.Bucket
```
 
## Notes

- Github Pages environment is currently broken because the baseURL is no longer set to `https://3b1b.github.io/3Blue1Brown.com/`  in config.


KurtBruns:
The current branch variable needs to be made available for creating Git LFS URLs in the **staging** environment. @3b1b Assuming Vercel works like the environment that runs the github action, try updating Vercel's build command to something like this:

```
env HUGO_PARAMS_BRANCH=`git rev-parse --abbrev-ref HEAD` hugo --environment staging
```

There may be another way to get the current branch name in Vercel, but as long as `git` is available on the command line, this should work fine. Keep any other flags we currently pass to the build command. Once this is done, the current branch can be accessed like this.

```
.Site.Branch
```

Then an asset URL in the staging environment can be build like the one below, where `workflow` is replaced with the current branch name.

```
https://github.com/3b1b/3Blue1Brown.com/raw/workflow/content/lessons/2021/git-lfs-test/example.mp4
```

KurtBruns:
It appears that Vercel is successfully passing the current branch name to Hugo. The current branch name can be accessed in the **staging** environment for templating like this:

```
{{ .Site.Branch }}
```

The Linode Bucket URL can be access in the **production** environment like this:

```
{{ .Site.Bucket }}
```

To simulate the staging environment, run the command

```sh
env HUGO_PARAMS_BRANCH=`git rev-parse --abbrev-ref HEAD` hugo --environment staging
```

To simulate the production environment, run the command

```
hugo --environment production
```

**Note:** The default environment is set to "development" when running `hugo server` and "production" when running just `hugo`. For our purposes, I think it makes sense to view the "development" environment as the same as "local". 

I recommend that instead of passing a bunch of flags to the **staging** and **production** environments that we define that configuration for each environment in its config file. For example, in the staging environment's config file `buildFuture` would be set to true instead of passing it in as a flag.

I'm going to log off for today. @vincerubinetti can you modify the figure shortcode to be environment aware? If you run into issues or otherwise, I can take a crack at it tomorrow morning.

KurtBruns:
The `figure-test` shortcode demonstrates environment aware URLs. See the `content/lessons/2021/git-lfs-test/` for an example. 

http://localhost:1313/lessons/git-lfs-test/

vincerubinetti:
Do you want to delete the test figure component, and put the real one into effect in this PR. I believe a clean way to do it would be to make `layouts/partials/asset-url.html`:

```
{{- $path := . -}}
{{- if $resource := .Page.Resources.GetMatch $path -}}
  {{- if (eq site.Params.environment "production") --}}
    {{- printf "%s%s" site.Params.Bucket $resource.RelPermalink ) --}} 
  {{- else if (eq site.Params.environment "staging") --}}
    {{- printf "https://github.com/3b1b/3Blue1Brown.com/raw/%s/content/%s/%s" site.Params.branch .Page.File.Dir $path --}}
  {{- else --}}
    {{- resource.RelPermalink --}}
  {{- end --}}
{{- else -}}
  {{- errorf (printf "%s: Unable to find resource given the path:\"%s\"" .Position $path) -}}
{{- end -}}
```

Then in `layouts/shortcodes/figure.html`, change `partial rel-url` mentions to `partial asset-url`

PR #50, opened 2021-05-26
Add basic text bubble for pi characters

vincerubinetti:
screenshots:

image

image


vincerubinetti:
> My guess is that these should only be used if there's like 1 or 2 words in them.  Maybe if any message has more words than that, just throw it above the pi creature with no bubble rather than worrying about any awkward bubble resizing logic.

I figured if you had more than a few words to say, you should just put it in the paragraph above or below? But if you want, we can make this component have a `caption` parameter as well to put text under and centered, like the figure component has. Then just leave it to the author to use either `caption` or `text`.

3b1b:
> I figured if you had more than a few words to say, you should just put it in the paragraph above or below?

Yes, that makes sense.  I think I was trying to poke at edge cases, but it would be silly to bake a case into the component which allows bad behavior.

vincerubinetti:
Also to do still: 
- add `thought` parameter. 
- tweak the speech and thought bubble SVGs @3b1b sent me such that they're roughly the same proportions, a little bit thicker, and their `viewBox` attributes are their text-safe areas like this:
image

Will do these when I get home. Probably will just do the thought bubble from scratch like I did with the speech bubble, taking care to match the branding (same look as the videos).

3b1b:
TBH, that thought bubble is a little wonky and I have zero attachment to it.

vincerubinetti:
Okay, ready for merge!

PR #51, opened 2021-05-28
incorp patreon patron list

### LINKED ISSUE #7: Patreon name list near lesson footer
Put list of Patreon supporters at bottom of each lesson. Those who contributed between a month leading up to the lesson's release?

Maybe pull from Patreon API? Or perhaps more simply, just make `_data/patrons.yaml` like:

```yaml
# a supporter
- name: Adrienne Wilson
  start: 2020-10-24
  end: 2018-08-10
  star: true


# another supporter
...
```

Maybe have a `special` field for people who have supported > 1 year or high donation amounts, and highlight them in bold or gold or something?

vincerubinetti:
Closes #7 

- rename `.yml` to `.yaml` per YAML's recommendation, rename `sync` workflow to `sync-assets` to be more descriptive. these shouldn't affect anything as far as i can tell, but @KurtBruns please take a look
- add some stuff to git ignore
- remove lesson archetype. it's not that useful
- blockquote appearance: lighter gray and more padding (can't go any lighter because there should be sufficient color contrast for hard-of-seeing readers)
- show patrons and discussions on all lessons, not just ones with text
- add expandable/collapsible list of patrons to patron partial reading from `patrons.txt`. 
- fix figure width/height params
- write node script to query patreon API to get list of all supporters, their names, their start/end dates, and their total lifetime contribution amount (i don't think there's a way to get how much a member was donating at a specific time. you can only get lifetime amount and current amount). the script gets all that, does name replacement, and exports it to `data/patrons.yaml`
- make github action to run the node script a couple times a month to keep the list up to date
- delete some "test" components
- create support page with basic links and wall of thanks, sorts by total lifetime contribution


vincerubinetti:
@3b1b the name overrides go in `scripts/get-patrons/name-overrides.yaml`, like this:

image

3b1b:
Looks pretty good.  Can we add a filter to have it just show those supporting at $16+, and then display it one name per line, multicolumn.

3b1b:
In a perfect world, the Patreon API would provide some data that enables a snapshot of who was supporting and at what amount on a given date, but from what I can tell doesn't seem quite possible.  Would you agree?

I do have the data on what names should be listed for each video, based on the "wall of thanks" pages linked to in the video descriptions, though it'd be a task to go through (interns...).  Do you think it might be better to store the supporters-to-thank information on a per-page basis, e.g. in the folder of the relevant lesson?

Maybe it makes sense to have a general "wall of thanks" somewhere on the site (using everything from this PR), which would list all the current supporters.  The bottom of each lesson page could then read something like "Special thanks to the individuals below for their support of this particular lesson, and further thanks to the [current patrons] (link-to-general-wall-of-thanks) for helping to fund future projects."

vincerubinetti:
That sounds reasonable to me. A few questions:

1. I guess it'd be easiest to just have the names in text file format, one name per line? Doesn't need to be .yaml, hugo can read .txt. Maybe you can provide one for testing this PR?
2. Where should we put this wall of thanks? My thought would be to just make a `support` page that has some text ("this channel is supported by your generosity", "patrons get store discounts"), some links ("become a patron", "$5", "$20", etc), then the lists. This would duplicate some info already on the patreon page itself though, and it would require more clicks for the user to get to donating. Other ideas: have it be a separate `thanks` page, and put the link either on the homepage (not in the header), or in the center of the footer.
3. On the wall of thanks, maybe it'd be cool to have two lists, current supporters (sorted by current donation) and alumni supporters (sorted by all-time donation)? Just thinking it might be nice to show some people who donated highly in the past but don't happen to currently donate.

3b1b:
> Maybe you can provide one for testing this PR?

Just added one for the quick-eigen video, I can try to go and add all the others sometime this weekend.

> Where should we put this wall of thanks? 

I'm okay with having a support page that is slightly redundant to the Patreon page, after all that [currently exists](https://www.3blue1brown.com/support).  If it has the same buttons for $2, $5, $10, etc., it wouldn't even be more clicks.  The other advantage is that it's a more natural place to list other ways to contribute, e.g. one-time donations via Paypal.

> On the wall of thanks, maybe it'd be cool to have two lists, current supporters (sorted by current donation) and alumni supporters (sorted by all-time donation)? Just thinking it might be nice to show some people who donated highly in the past but don't happen to currently donate.

That makes sense to me, let's do it.

I realize I'm probably making this all tricker than it needs to me, thanks for the help.

PullJosh:
Per the request on Discord, I created `patrons.txt` files for all the old videos (at least, those which had lists available). I didn't realize the patrons branch already existed so those files are on the [patron-thanks](https://github.com/3b1b/3Blue1Brown.com/tree/patron-thanks) branch. Hopefully it's not too much trouble to merge those changes in.

vincerubinetti:
Updated. Added a basic support page with some links, and two lists of current and past supporters (sorted by donation amount). Lesson template now reads from `patrons.txt` in the same folder, and also shuffles the names on page load so the people with "A" names don't always show.

vincerubinetti:
Going to merge this so I can work on the site tonight. Assuming people are not around to review the last commit. I'll make sure to get Grant's feedback on the patreon lists later.

PR #58, opened 2021-06-02
Essence of Calculus Chapter 1

KurtBruns:
I'm opening this pull request to test Netlify's comment feature and roundtrip the markdownification process. There might be a couple more iterations to do on this lesson.


KurtBruns:
When writing a question, I think it might make sense to have full markdown support for the question. @3b1b any thoughts on this?

This would make authoring the question a little backwards, but the benefit would be that you could add footnotes, links and images in the question.

Before:

```md
{{<question
  question="What is the height of the rectangle whose left corner touches $r=0.3$? **Note:** This is the rectangle highlighted in the figure above."
  choice1="$6 \pi$"
  choice2="$\frac{3}{10} \pi$"
  choice3="$2 \pi$"
  choice4="$\frac{3}{5} \pi$"
  correct=4
>}}
```

After:

```md
{{<question
  choice1="$6 \pi$"
  choice2="$\frac{3}{10} \pi$"
  choice3="$2 \pi$"
  choice4="$\frac{3}{5} \pi$"
  correct=4
>}}

What is the height of the rectangle whose left corner touches $r=0.3$? **Note:** This is the rectangle highlighted in the figure above.

{{</question>}}
```

Alternatively, you could pass the input to the `question="..."` parameter to the [markdownify function](https://gohugo.io/functions/markdownify/).

Here is a screenshot of the example question:

Screen Shot 2021-06-03 at 11 00 46 AM

3b1b:
This makes a lot of sense to me. @vincerubinetti, would you agree?

vincerubinetti:
That's fine with me. Though you may want to save the `.Inner` for the explanation per #60 . Whichever you think is going to be typically longer. Then use the markdownify on the other.

PR #60, opened 2021-06-03
256 Bit Security Article

riverliway:
Markdownified the 256 Bit Security Article.
Modified the question shortcode to optionally show an explanation after the user is correct.
Created new dropdown shortcode (prototype) which shows/hides content when clicked. Further visual design is recommended.
Added new miner PI creature


3b1b:
I ended up making a few more edits. Upon a second look, the intro felt too much like it was beginning an article about hash functions, rather than just using them as an illustration of something where a huge about of guessing and checking is needed to break some security assumption. No doubt I'm overthinking it a bit for what is meant to be a simple side-dish style article.

I decided to make the die roll question a multiple-choice, and offer the explanation in the question as per your idea for adding explanations to that component. It's a little on the long side, but it felt unsatisfying not to explain it properly or show the math. For succinctness, then, I ended up removing the code example.

On the whole, this looks great, and I think your additions made it much better.

3b1b:
@vincerubinetti, before I merge this, can you look at the explanation addition in the question component and let me know if that seems reasonable?  There's probably a better way to style it to stand out as distinct from the article body a little more.

PR #67, opened 2021-06-04
adding draft of fourier chapter

leios:
This *should be* a translation of the Fourier Transform chapter from the old Honkit framework to Hugo. I know we are switching to Next next, but I figured I would put this up now.


riverliway:
I really think some sort of interactive demo would be a great addition for this article. I added an image showing how one such demo could work: it allows the user to change the input frequency and winding frequency which shows how the foundations of the machine work. The only thing is this demo doesn't allow for the sum of waves, which is a core use case for the Fourier transform. I'm open to hearing other designs which would allow the user to control the superposition of waves too.

3b1b:
Covered by https://github.com/3b1b/3Blue1Brown.com/pull/186

PR #72, opened 2021-06-13
Next port and more

### LINKED ISSUE #19: "Aside" component
Need some kind of footnote or aside component. There are several ways to do this.

One is by having a tooltip, but this is difficult and a bit hacky to do. Another is by having an accordion the reader can expand/collapse, but this, design wise, only makes sense for non-mid-sentence asides (it'd have to go after a full paragraph).

A perhaps temporary solution we came up with is: have a superscript icon or number that when clicked on expands a hidden paragraph below the current paragraph with rich content.

Need to experiment with this to figure out what the best doable solution is.

### LINKED ISSUE #46: Hugo Errors
I'm opening this issue about error reporting in Hugo with the goal to prevent silent failures and generate more meaningful errors for authors and developers while working with Hugo. This can be pretty subtle when managing the dot context, but when setup correctly it's no big deal and will make our lives better.

TLDR: Give more useful errors to content authors and prevent silent failures when content changes.

## General Example

As a general example, this *example* figure shortcode matches a page resource (image) based on a path supplied by the user.  For example, the user might write the shortcode shown below in their markdown file.

```
{{ figure image="example.png" }}`
```

The shortcode template looks like this:

```
{{ with .Page.Resources.GetMatch $path }}
  {{/* TODO: render here */}}
{{ else }}
  {{ errorf (printf "%s: Unable to find resource given the path:\"%s\"" .Position $path) }}
{{ end }}
```

Then, when the author types a bad file path they'll get a shorter, better error. Specifically, the line number where the failure occurred and the input that caused the error.

```
ERROR 2021/05/25 10:27:29 "~/Sites/3Blue1Brown.com/content/lessons/2021-05-22/index.md:12:1": Unable to find resource given the path:"bad-file-path.png"
```

There are a number of other considerations for the actual *figure* shortcode. For example, what happens if the "image" or "video" attribute is misspelled or not present.

## Silent Failures

In addition to making the error messages more helpful, managing errors and [warnings](https://gohugo.io/functions/errorf/) ourselves gives us the added benefit of preventing *silent* failures when editing the site. For example, the lesson gallery shortcode is a currently a good example, because if you change data in the `data/featured.yaml` file or, just as likely, change the path of a lesson in the `content` directory, the current behavior is silent failure - the bad id fails silently and no lesson card is rendered.

Currently on line 34-36 in `layouts/partials/lesson-gallery.html` the lesson card partial is used. If a bad id is passed to the partial, this renders nothing when it probably should report an error.

```
{{ range $featured := site.Data.featured }}
  {{ partial "lesson-card" (dict "id" .) }}
{{ end }}
```

An error can be thrown in the partial, and probably should throw one if the partial expects and string id and calls `GetPage`. However, I would recommend that we provide a more better error message here and instead pass the lesson page context and additional values to the partial.

```
{{ range site.Data.featured }}
  {{ with site.GetPage . }}
    {{ partial "lesson-card" (dict "context" .) }}
  {{ else }}
    {{ errorf (printf "data/featured.yaml: Unable to find the lesson associated with the id:%s" . )}}
  {{ end }}
{{ end }}
```

The same applies for the `data/topics.yaml` file. For example, before the last commit, adding this check to lines 43-45 in `layouts/partials/lesson-gallery.html`  reports the three errors shown below while testing.

```
ERROR 2021/05/25 13:57:42 data/topics.yaml: Unable to find the lesson associated with the id: "quick-eigen"
ERROR 2021/05/25 13:57:42 data/topics.yaml: Unable to find the lesson associated with the id: "qa1"
ERROR 2021/05/25 13:57:42 data/topics.yaml: Unable to find the lesson associated with the id: "qa2"
```

## Summary

In general, I agree with Vince that first priority is errors that content authors run into, usually associated with shortcodes, data files and adding/editing lessons in the `content` directory. If we run into an obscure error in the future, we can always review it and see if we can provide a better one.

### LINKED ISSUE #55: Chips for text-version and other things
We have chips for chapter coming in the next PR. It'd be good to have chips for other important things like, "this lesson has a text version ready", and less importantly, "this lesson has interactive elements".

### LINKED ISSUE #61: Option for figure captions to change based on view mode
Unless I am mistaken, the figure caption remains the same for both the Still and Animation tabs, but it makes sense to have a separate caption for certain cases where the still needs a bit more explanation.

### LINKED ISSUE #62: Width of internal link preview
The tooltip preview of internal links sort of runs awkwardly against the edge when the description is long. I think we should decrease the maximum width it can have, and also give it some buffer against the edge of the screen.

image

### LINKED ISSUE #63: Question component explanation
As per https://github.com/3b1b/3Blue1Brown.com/pull/60, and also https://github.com/3b1b/3Blue1Brown.com/pull/58, the question component should have an explanation, and this explanation should have some styling that makes it stand out clearly from the article.  I'm guessing lower width and clearly distinct background color?

The explanation is likely to be larger than the question, and if we have to choose which one gets full markdown, it should be the explanation.

Separate question: Should "show explanation" be a separate operation from "check answer"?

### LINKED ISSUE #66: Table of contents tweaks
Some thoughts on the table of contents. These are mostly (entirely?) shamelessly lifted from the design of the sidebar on Wikiwand.

- As soon as one scrolls down to the article, the table of contents should appear on the right by default.
   - But only if the screen is wide enough, of course.
   - As a side note, I think it might look nice to have the width of the article content a bit thinner, maybe around 900px instead of 1080px
- Instead of saying "Lesson sections" on the top it should either be nothing, or the title of the article.
- The name of the section where the reader current is should be highlighted (or somehow indicated) in the sidebar, so as to give an at-a-glance sense of how far in the lesson they sit.
- (Nice to have, not critical) in clicking to a new section, rather than jumping to it there should be a quick animated scroll to the appropriate position.

vincerubinetti:
closes #66
closes #63 
closes #62 
closes #61 
closes #55 
closes #46 
closes #19

- port everything to next.js
- improve table of contents auto behavior and appearance
- improve accordion component
- add generic tooltip component
- add separate captions for figure image and video
- add generic "markdownify" component that can make any component parameter accept markdown
- add chip for "does this lesson have a text version"
- add explanation to question component
- other various tweaks and fixes


vincerubinetti:
Before reviewing code, let us meet tomorrow quickly and discuss.

Regarding performance, it improved quite noticeably, at least on my computer. I believe this is probably due to the changes I made to `util/pages.js`, where I tried to pre-compute and only compute once as much as possible, rather than doing it each time a page is requested.

vincerubinetti:
Okay the former "to-dos" in this PR have been captured in https://github.com/3b1b/3Blue1Brown.com/issues/73.

Grant feel free to merge once merge conflict is resolved and the rest can be fixed later.

vincerubinetti:
Here's a brief discussion on my thoughts on Jekyll/Hugo (SSG) vs Next, having worked fairly extensively in all three now.

SSG are a bit simpler to use and learn, because it's just vanilla html/css/js, with a little bit of custom templating language thrown in (which Jekyll's is better than Hugo's). With Next, there are many more technologies involved, including node, webpack, react, mdx, jsx.

SSGs can also be faster, I think, because they make more assumptions about the type of content and where it is loaded. SSGs, for example, automatically compile and give you access to a data structure of all pages on the site that you can loop through for any purpose. With next, you have to build this list yourself, and thus it's not cached out of the box.

Next is still probably a cleaner way to go forward though. For one, the javascript, css, and html for a component is all co-located in one folder. In SSGs, you have a disconnect between the component (include/partial in jekyll/hugo) and the javascript for interactivity. In the js, you have to go find the proper element in the document and make modifications to it. This is fragile to changes in the html structure or even just the html element class names and ids. In react, this is all co-located and more declarative. React also avoid more global state, as again, html and js are co-located in the same function scope.

Next is also less prescriptive of where you put things, so for the most part, you can just import things like you would do in any programming language. Want to include a component in another component? No problem. In hugo, you couldn't include a partial directly in markdown, and you couldn't use a shortcode in a partial. In react, everything's just a function/component, making it simpler. 

Next will also make it cleaner and more robust to include interactive react apps in the site. With SSGs you either would have to use an iframe (which sandboxes the app and you can't style from the outside), or some more complicated solution where the app is built and its compiled result is included in the page and mounted to a unique container.

Overall, I think Next will be the right choice going forward.

There's probably many other things to say but I can't remember them right now. Others can chime in if they want to. This is just for posterity anyway.

PR #77, opened 2021-06-14
fix figure src, querySel to getElById, markdown question, file and dir page props

vincerubinetti:
- include original loaded file (from `/public`) path and directory in page props/metadata
- make figure prepend proper directory to relative image/video sources
- change certain `querySelector`s to `getElementById`s to avoid bug where a heading starts with a number and thus the selector becomes e.g. `#1-in-4-billion-success-rate`, which is not a valid CSS selector
- make question text support markdown
- other minor fixes


vincerubinetti:
Superceded by #78

PR #78, opened 2021-06-14
Various fixes

vincerubinetti:
- include original loaded file (from `/public`) path and directory in page props/metadata
- make figure prepend proper directory to relative image/video sources
- change certain `querySelector`s to `getElementById`s to avoid bug where a heading starts with a number and thus the selector becomes e.g. `#1-in-4-billion-success-rate`, which is not a valid CSS selector
- make question text support markdown
- remove lowercase components from git index
- fix math backslash escape issue?
- intelligently wrap markdown in section
- scroll view to topic card when you open it
- other minor fixes


vincerubinetti:
Going to merge this as I think it's causing problems. I have the whole repo backed up in multiple places just in case we can't revert for some weird git reason.

PR #82, opened 2021-06-15
Lesson: Differential Equations, Chapter 5

vivek3141:
Pull request checklist:

- [ X ] I have requested a review from the appropriate persons or have followed the appropriate review processes
- [ X ] I have checked that my changes work in the preview
- [ X ] I have checked that the rest of the site is still functioning in the preview
- [ X ] I have checked that my content/code is consistent with existing content/code
- [ X ] I have followed the guidelines in the readme

This pull request: Adding the CH5 Lesson


3b1b:
It's possible you weren't done and I'm taking a premature look, but I thought I'd share a few comments.

- Did you mean to include the file anims.py?
- I can't see this animation:
image
- We could make the phrase "next lesson" a rich link with `<LessonLink id="matrix-exponents">next lesson</LessonLink>`
- I would remove the initial condition $e^0 = 1$ from these two images. 
image
image
- This question seems to have no options. 
image
-

vivek3141:
I've fixed the issues you mentioned. I believe the syntax of the question object was changed, which caused it to break. The others were a result of me not cross-checking after Kurt helped me fix the git issues. Let me know if it looks good, then we can merge it.

3b1b:
Okay, at the risk of going back to this one too many times, after one closer look I do have more suggestions. This all should have been in my initial review, but after this, I promise to let this one go :)

- Provide animations behind the images 1_img, 2_img, and -0.5_img, since these are intrinsically dynamic. I might even make those animations the default view, but I leave it up to you which one makes for a better reading experience.
- In the sentence "So after pi seconds, you’ve traced a distance of pi around", use the symbol $\pi$, since that's consistent with how it shows up on the next line with $\tau = 2\pi$.
- Add some section titles, even though it's a small article.  Maybe "The real case", "The complex case" and then (as per the next point), "But what is it?"
- (Optional), rewrite the last paragraph to expand on the fact that when I wrote "immoral", what I really meant was raising a number to a complex value might still feel like it makes no sense, at least if you're thinking of it as meaning "e multiplied by itself πi times".  Mention that for complex inputs, the definition of $e^x$ is $1 + x + \frac{x^2}{2} + \frac{x^3}{6} + \cdots + \frac{x^n}{n!} + \cdots$.  You could even try to incorporate [this animation](https://www.dropbox.com/s/bgjacalkgk28rho/ExpMeaningWithoutCircle.mp4?dl=0) to show what the content of the statement $e^{\pi i}$ really is.

PR #93, opened 2021-06-17
Sync Action Update

kurtbruns:
Opening this pull request to preview and test the sync action.

Changes:

- Changed `.gitattributes` rules to reference `public/content/**/...` instead of `pages/lessons/**/...` after noticing lesson media files not managed by LFS.
- Added `dir` when building the bucket URL
- Excluded SVG file types from be prepended with bucket URL since they are served from Netlify

@3b1b I tested the production environment locally and things looked fine. Next step would be to merge into main and check the live site. As long as the `process.env.mode` is set to `production` it should work as intented.


3b1b:
Outstanding, thanks!

PR #109, opened 2021-06-22
Windmills

vivek3141:

- [X] Includes authorship line: `Text adaptation by [Name]`
- [X] The lesson has an informative/compelling description, fit for a preview on the home page.
- [X] The lesson has no grammar/spelling errors
- [X] The lesson has no mathematical errors
- [X] The table of contents effectively conveys the overall structure of the lesson
- [X] The lesson uses no `# Top Level` headings
- [x] The lesson includes well-chosen comprehension questions

- I couldn't think of places where comprehension questions would be helpful to the learning process in this article

- [X] The lesson incorporates the core visuals from the video
- [X] The visuals are clear and helpful for understanding
- [X] The components are used in the places and ways they are intended (see the readme)


public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+    image="usa_imo.png"
+/>
+
+The contest is essentially a test, split over two days, with three questions given over four and a half hours each day. The questions are all proofs, meaning you don’t simply find some numerical answer, you have to discover and articulate a rigorous line of reasoning to answer each difficult question, and they are scored on a scale from 0 to 7.
+
+<Figure
+    image="imo_q.png"
+/>
+
+Of interest to us today is the one from 2011, with 563 total participants representing 101 countries. I know what you’re thinking, and the answer is yes. Those do all happen to be prime numbers. But that’s not why this test was interesting, out of these prime problem solvers, only one, Lisa Saurmann from Germany, got a perfect score. The only thing standing between the next two runners up and a perfect score that year was problem #2. This problem is beautiful, and despite evading many of the world’s best mathematicians of their age, the solution is something which anyone watching this video can understand.
```
3b1b, line 32: Pull out the line "I know what you’re thinking, and the answer is yes. Those do all happen to be prime numbers. But that’s not why this test was interesting," to be a footnote.

Then just start the next sentence with "Out of these participants, only one..."

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+
+## The problem
+
+So let’s begin by reading it carefully.
+
+> Let S be a finite set of at least two points in the plane
+
+Mmkay, as you read a question it’s often helpful to start drawing out an example for yourself.
```
3b1b, line 40:
```suggestion
As you read a question it’s often helpful to start drawing out an example for yourself.
```

Feel free to remove words/comments that are meant more as notes for reading it out loud.

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+The contest is essentially a test, split over two days, with three questions given over four and a half hours each day. The questions are all proofs, meaning you don’t simply find some numerical answer, you have to discover and articulate a rigorous line of reasoning to answer each difficult question, and they are scored on a scale from 0 to 7.
+
+<Figure
+    image="imo_q.png"
+/>
+
+Of interest to us today is the one from 2011, with 563 total participants representing 101 countries. I know what you’re thinking, and the answer is yes. Those do all happen to be prime numbers. But that’s not why this test was interesting, out of these prime problem solvers, only one, Lisa Saurmann from Germany, got a perfect score. The only thing standing between the next two runners up and a perfect score that year was problem #2. This problem is beautiful, and despite evading many of the world’s best mathematicians of their age, the solution is something which anyone watching this video can understand.
+
+## The problem
+
```
3b1b, line 35: I would start by showing the full question, then breaking it down piece by piece.

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+    image="intro_img3.png"
+    video="intro_img3.mp4"
+    show="video"
+/>
+
+> This point, Q, takes over as the new pivot, and the line now rotates clockwise about Q, until it next meets a point of S. This process continues indefinitely.
+
+<Figure
+    image="intro_img4.png"
+    video="intro_img4.mp4"
```
3b1b, line 68: This one might be satisfying if we found a clip that can loop, then we set loop={true} in the component.

Perhaps the section from roughly 0:25 - 0:35 of this: https://www.dropbox.com/home/3Blue1Brown/videos/2019/windmill/staged_scenes_0?preview=Scene_007_IntroduceWindmill.mp4

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+<Figure
+    image="intro_img3.png"
+    video="intro_img3.mp4"
+    show="video"
+/>
+
+> This point, Q, takes over as the new pivot, and the line now rotates clockwise about Q, until it next meets a point of S. This process continues indefinitely.
+
+<Figure
+    image="intro_img4.png"
```
3b1b, line 67: For the still, I'd add a little more space between the left and right parts to make clear that they are different.  Maybe even add a thin stroke width box around each.

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+    show="video"
+/>
+
+Alright, that’s kind of fun. We keep rotating and changing the pivot. You can see why they call it a windmill process, and you can also see why they specified that no three points lie on one line. You wouldn’t want to run into the ambiguity where you don’t know what pivot to switch to. Okay, so what’s the question?
+
+> Show that we can choose a point P in S and a line l going through P such that the resulting windmill uses each point of S as a pivot infinitely many times.
+
+<Figure
+    image="intro_img5.png"
+    video="intro_img5.mp4"
```
3b1b, line 78: I know I said ~10 seconds for clips, but this one might benefit from going on long enough for someone to be convinced that all the points will be hit infinitely many times.

3b1b: One possibility would be to speed [this clip](https://www.dropbox.com/home/3Blue1Brown/videos/2019/windmill/1440p60?preview=WindmillExample15Points.mp4) by 2x, or even 3x:

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+    show="video"
+/>
+
+## Why it’s “unusual”
+
+Depending on your tolerance of puzzles for puzzles’ sake, you might wonder why anyone would care about such a question. There’s a very good reason: I’d argue that the act of solving this will make you better at math and other related fields, which I’ll explain once you’ve seen the solution.
+
+But certainly on the surface this feels disconnected from other parts of math. Other Olympiad problems involve some function to analyze, or a numerical pattern to deduce; or perhaps a difficult counting setup or an elaborate geometric construction, but problem 2 is an unusually pure puzzle. In some ways, that’s its charm: proving that some initial condition will result in this windmill hitting all points infinitely many times doesn’t test if you know a particular theorem, it tests if you can find a clever perspective.
```
3b1b, line 86:
```suggestion
But certainly, on the surface, this feels disconnected from other parts of math. Other Olympiad problems involve some function to analyze, or a numerical pattern to deduce, or perhaps a difficult counting setup or an elaborate geometric construction. But problem 2 is an unusually pure puzzle.  In some ways, that’s its charm. Proving that some initial condition will result in this windmill hitting all points infinitely many times doesn’t test if you know a particular theorem, it tests if you can find a clever perspective.
```

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+
+But that blade cuts both ways; without resting existing results from math, what could someone possibly study that would prepare them to solve this? In fact, this brings us to the second unusual thing about this problem: Based on the results, I’m guessing it turned out to be much harder than the contest organizers expected.
+
+You see, typically, the three problems in each day are supposed to get progressively harder. They’re all hard, of course, but problems 1 and 4 should be doable, problems 2 and 5 should be challenging, and problems 3 and 6...well they can be brutal.
+
+But take a look at how many of the 563 participants that year got perfect scores on each problem: Only 22 got a perfect score for this question number 2. By contrast, 170 got a perfect score on problem 5, which is supposed to be about the same difficulty, and more than twice as many, got a perfect score for problem 3, which is supposed to be harder.
```
3b1b, line 92: Include a graphic.  Especially if we're using the phrase "Take a look"

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+
+You might notice that only six students got full points for problem 6 that year, so by some measure that was the hardest problem on the test. In fact, the way I introduced things earlier was a little disingenuous, the full data would suggest problem 6 was the real clencher. But what’s strange is if you look at the results of the only six students to solve problem 6, all of whom are clearly phenomenal, world-class problem solvers, this windmill puzzle evaded five out of six of them.
+
+## Trying things out
+
+But again, this problem is hard not hard because of the background knowledge it demands, it asks only for insight. So how do you approach something like this?
```
3b1b, line 98:
```suggestion
Again, this problem is not hard because of the background knowledge it demands. It asks only for insight. So how do you approach something like this?
```

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+
+The first step with any puzzle like this is simply to play around to get a feel for it, and it’s always good to start simple and slowly get more complicated from there.
+
+The simplest case would be two points, where the line trades off between each point, so that works well enough.
+
+<Figure
+    image="2_img.png"
+    video="2_img.mp4"
```
3b1b, line 106: Make into a loop.

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+    show="video"
+/>
+
+Adding a third point, it’s pretty clear the line will just rotate around all of them. It might not be entirely clear how you’d phrase that as a rigorous proof yet, but right now you’re just getting a feel for things.
+
+<Figure
+    image="3_img.png"
+    video="3_img.mp4"
```
3b1b, line 114: Likewise, make a loop.

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+## The solution
+
+As a general problem-solving tip, whenever you have a vague idea that feels productive, find a way to be more exact about what you mean, preferably putting numbers to it, then ask questions about those numbers.
+
+In our example, one way to formalize the idea of a “middle” is to count how many points are on either side of the line. If you give the line some orientation, you can reasonably talk about the left half, say coloring all points to the left of it blue, and the right half, coloring all points to the right of it brown, and what it means for the line to be in the middle is that there are as many blue points and brown points.
+
+For the moment, let’s say the total number of points is odd, and the point the line passes through is colored white. So for example, if there were 11 points, then you’d have five blue ones on the left of the line, and five brown ones on the right of the line, and the one white one at the pivot. The case with an even number of points will be similar, just slightly less symmetric.
+
+<Figure
+    image="introduce_colors.png"
```
3b1b, line 162: If we describe the point the line passes through as white, we should make sure it's white in this image.

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+
+<Figure
+    image="change_pivot.png"
+    video="change_pivot.mp4"
+/>
+
+Great, that’s insight #1. Now why would this imply that the line must hit every point infinitely many times, no matter what weird set of points you dream up? Well, think about letting this process go until the line has turned 180 degrees. That means it’s parallel to its starting position, and because it has to remain the case that half the points are on one side, half are on the other, it must be passing through the same point it started on. Otherwise, if it ended up on some other point, it would change the number on a given side. Additionally, since the line has rotated halfway around, everything that was blue has become brown, and everything which was brown has become blue, and the only way to change color is if you get hit by the line.
```
3b1b, line 180:
```suggestion
Great, that’s insight #1. Now why would this imply that the line must hit every point infinitely many times, no matter what weird set of points you dream up? Well, think about letting this process go until the line has turned $180^\circ$. That means it’s parallel to its starting position, and because it has to remain the case that half the points are on one side, half are on the other, it must be passing through the same point it started on. Otherwise, if it ended up on some other point, it would change the number on a given side. Additionally, since the line has rotated halfway around, everything that was blue has become brown, and everything which was brown has become blue, and the only way to change color is if you get hit by the line.
```

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+
+So for our odd number case, that means that after a half rotation, the line is back to where it started, having hit all the other points. So as time moves forward, it repeats that exact same set of motions over and over, hitting all points infinitely many times.
+
+If there are an even number of points, we need to alter the scheme slightly, but only slightly. Let’s say that in that case, the pivot counts as a brown point. So we can still select an initial point so that half the points are blue, all on the left, and half are brown, now either meaning their either on the right or the pivot.
+
+<Figure
+    image="even_intro.png"
+/>
+
+The same argument applies that after a 180-degree turn, all the points will have swapped color, although this time the line might be passing through a different point after that first half turn, specifically one that used to be blue. But after another 180 degrees the line must be passing through the same point it started on. Why? Again, it must be parallel to its starting position, and if it was passing through any other point, the number of points on a given side would be different.
```
3b1b, line 196:
```suggestion
The same argument applies that after a $180^\circ$ turn, all the points will have swapped color, although this time the line might be passing through a different point after that first half turn, specifically one that used to be blue. But after another $180^\circ$ the line must be passing through the same point it started on. Why? Again, it must be parallel to its starting position, and if it was passing through any other point, the number of points on a given side would be different.
```

public/content/lessons/2019/windmills/index.mdx
3b1b: This is also a point where a question could be inserted.

<Question
  question="If the line has rotated $180^\circ$, and ends up passing through the same point where it began, has it necessarily touched every point along the way?"
  choice1="Yes"
  choice2="No"
  answer={1}
>

Since the line has rotated halfway around, every point which was initially to the line's left, and so was colored blue, is no to the line's right, and so is colored brown. Likewise, every point which began brown becomes blue. The only way to change color is if you get hit by the line.

</Question>

public/content/lessons/2019/windmills/index.mdx
3b1b, line 196: If you wanted, this is another opportunity for a question.  Ask whether in this even-number case, after a $180^\circ$ turn, will the line be passing through the same point where it began.

public/content/lessons/2019/windmills/index.mdx
```diff
@@ -1,9 +1,232 @@
...
+
+<Figure
+    image="even_img.png"
+    video="even_img.mp4"
+    show="video"
+/>
+
+So again, you have a cycle which hits all points, and which ends in the same position where it started, so it must hit all points infinitely many times.
+
```
3b1b, line 205: Below is something we could insert here. Maybe as an accordion?

---

From the comments on the original video for this topic, I learned that many people didn't quite realize that the problem had been solved at this point. Here's what I wrote in response:

> A few people ask "but what is the formal solution?".  Take a look at [this writeup](https://artofproblemsolving.com/wiki/index.php?title=2011_IMO_Problems/Problem_2), which you'll find presents an argument with the same essential idea as was presented here.  It would get you full marks on the test: 
> 
> The fact that an argument doesn't reference equations and sets doesn't necessarily make it incomplete.  Sure, you could explicitly define what words like "left" and "right" refer to by defining such and such cross product with such and such parameterization of the line, but for most readers, it wouldn't actually remove any ambiguity.  The purpose of formality is to make all the terms used unambiguous to all readers, not to dress up the language to involve sufficiently many symbols. If you understood this video, you understood the full solution.

PR #134, opened 2021-06-25
Essence of Calculus Chapter 3

kurtbruns:
[Netlify PR Preview Lesson Link (Part 1, Power Rule)](https://deploy-preview-134--3blue1brown.netlify.app/lessons/derivatives-power-rule)
[Netlify PR Preview Lesson Link (Part 2, Trig Functions)](https://deploy-preview-134--3blue1brown.netlify.app/lessons/derivatives-trig-functions)

Peer review checklist:

- [ ] Includes authorship line: `Text adaptation by [Name]`
- [ ] The lesson has an informative/compelling description, fit for a preview on the home page.
- [ ] The lesson has no grammar/spelling errors
- [ ] The lesson has no mathematical errors
- [ ] The table of contents effectively conveys the overall structure of the lesson
- [ ] The lesson uses no `# Top Level` headings
- [ ] The components are used in the places and ways they are intended (see the readme)

Notes for Peer Review:

- I think the opening quote has the potential to be off-putting for students who haven't decided whether they love or hate math. Possibly replace with another quote?
- Plan to split lesson into two parts: 1) Derivatives of Monomial Terms and 2) Derivatives of Trig. Functions
- For the derivative of 1/x, the answer is left out of the video to encourage the viewer to engage with the content. For the article, should the solution be included in an accordion?

This pull request: 

- Publishes Essence of Calculus Chapter 3
- Split the lesson into two chapters: power rule and trigonometric functions
- Updates lesson layout and video component to allow for a timestamp to be specified in `.mdx` front matter

Thoughts about lesson splitting:

Pretty straightforward if one chapter becomes two chapters. Requires updating the chapter number in the front matter and the `id` data  in `topics.yaml`, personally I would define series structure in one or the other.

Probably makes sense to change the lesson thumbnails, where are thumbnails being pulled in from?


kurtbruns:
[Netlify PR Preview Lesson Link](https://deploy-preview-134--3blue1brown.netlify.app/lessons/derivative-formulas-geometrically)

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+  show="video" 
+/>
+
+Graphical intuition with slope can tell us why this derivative is high on the left, $0$ at the origin, and high on the right, but just thinking in terms of graphs would not land us on the precise quantity $3x^2$. For that, we had to take a much more direct look at the actual meaning of the function.
+
+### Derivative of $f(x) = x^n$
+
+<!-- TODO: incorporate text into expressions-->
+
+In practice, you wouldn’t necessarily think of the square every time you’re taking a derivative of $x^2$, nor would you necessarily think of the cube when taking a derivative of $x^3$. Instead, thinking like a mathematician, can we generalize this approach to see if a pattern emerges? And maybe, if we are lucky, we invent a tool that we can use to find the derivative of any old polynomial.
```
3b1b, line 137:
```suggestion
In practice, you wouldn’t necessarily think of the square every time you’re taking a derivative of $x^2$, nor would you necessarily think of the cube when taking a derivative of $x^3$. Instead, thinking like a mathematician, can you generalize this approach to see if a pattern emerges? Can you invent a tool to find the derivative of any polynomial?
```

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+<Question
+  question="What is the derivative of the function $f(x)= 3x^2$?"
+  choice1="$2x$"
+  choice2="$5x$"
+  choice3="$3x$"
+  choice4="$6x$"
+  answer={4}
+>
+
+This question can be modeled and solved as a question of change in area. For example, nudging the input of the function by $dx$ gives the expression $3(x+d x)^{2}$ which expands to the expression $3(x + dx)( x + dx)$. I've drawn the area below where the $3$ scales one of the sides.
```
3b1b, line 180:
```suggestion
You certainly don't need to visualize this to answer the question. In fact, in practice, it would be absurd to do so. Nevertheless, continuing with the playful and visual spirit of this lesson, it's fun to think of a few ways to do just that.

This question can be modeled and solved as a question of change in area. For example, nudging the input of the function by $dx$ gives the expression $3(x+d x)^{2}$ which expands to the expression $3(x + dx)( x + dx)$. I've drawn the area below where the $3$ scales one of the sides.
```

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+
+<Figure
+  image="figures/power-rule-example.svg"
```
3b1b, line 183: I might opt for the 3 copies of the (square + nudged area) setup, just because it better emphasizes why multiplying the function by 3 just multiplied the derivative by 3.

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+
+Looking at the figure we can see the rate of the change of the function represented by the new area of $6$ thin rectangles with area $x \cdot dx$. Alternatively, the question can be modeled as volume, where instead of scaling one of the sides, the $3$ extends the geometry into another dimension.
+
+<Figure
+  image="figures/3x^2-volume-model.svg" 
+  caption="Derivative of $f(x) = 3x^{2}$ modeled as volume"
+/>
+
+This gives us the equivalent derivative, as the new volume of the two rectangular prisms is equal to $3 \cdot x \cdot dx + 3 \cdot x \cdot dx = 6x$. As before, we are ignoring the smaller rectangular prism because its volume $3dx^2$ contains $dx^2$.
+
```
3b1b, line 195:
```suggestion

In either case, the upshot is that multiplying a function by $3$ also multiplies its derivative by $3$. And of course, this goes for any constant, not just $3$. In practice, this fact is simple enough to memorize, and you'll use it thousands of times in your relationship with calculus, but every now and then it's nice to step back and reflect on why it's true and what it means in a particular context.

```

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+</Question>
+
+### Derivative of $f(x) = \frac{1}{x}$
+
+As another example, think of the function $f(x) = \frac{1}{x}$. Now, on the one hand, you could blindly try applying the power rule, since $\frac{1}{x}$ is the same as writing $x^{-1}$. That would involve letting that $-1$ hop down, leaving behind one less than itself, $-2$.
```
3b1b, line 200:
```suggestion
As another example, think of the function $f(x) = \frac{1}{x}$. Now, on the one hand, you could blindly try applying the power rule, since $\frac{1}{x}$ is the same as writing $x^{-1}$. That would involve letting that $-1$ hop down to become a coefficient, leaving behind one less than itself in the exponent, $-2$. 
```

3b1b: Would it be worth adding a small illustration of what's meant by "hopping down" here? Or is it pretty clear?

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+This gives us the equivalent derivative, as the new volume of the two rectangular prisms is equal to $3 \cdot x \cdot dx + 3 \cdot x \cdot dx = 6x$. As before, we are ignoring the smaller rectangular prism because its volume $3dx^2$ contains $dx^2$.
+
+</Question>
+
+### Derivative of $f(x) = \frac{1}{x}$
+
+As another example, think of the function $f(x) = \frac{1}{x}$. Now, on the one hand, you could blindly try applying the power rule, since $\frac{1}{x}$ is the same as writing $x^{-1}$. That would involve letting that $-1$ hop down, leaving behind one less than itself, $-2$. 
+But let’s have some fun and see if we can reasonable this geometrically, rather than just plugging it through a formula.
+
+The value $1/x$ is asking “what number multiplied by $x$ equals $1$”, so here’s how I’d visualize it: Imagine a little rectangular puddle of water in two dimensions with area $1$. Let’s say that it’s width is $x$, which means the other must be $1/x$, since the total area is $1$.
```
3b1b, line 203:
```suggestion
The value $1/x$ is asking “what number multiplied by $x$ equals $1$”, so here’s how I’d visualize it: Imagine a little rectangular puddle of water in two dimensions with area $1$. Let’s say that it’s width is $x$, which means its height must be $1/x$, since the total area is $1$.
```

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+For the derivative, imagine nudging the input $x$ up by a value $dx$. How must the height of this rectangle change so that the area remains unchanged at $1$? That is, increasing the width by $dx$ adds some new area to the right here, so the puddle must decrease in height by some $d(1/x)$ so that the area lost off the top here cancels that out.
+
+<Figure
+  image="figures/11.41.svg" 
+/>
+
+You should think of that $d(1/x)$ as being some tiny *negative* value, since it’s *decreasing* the height of this rectangle. And you know what? I’m gonna leave the last few steps here for you to pause and ponder. And once you work out $d(1/x)/dx$, compare it to what happens if you apply the power rule purely symbolically to $x^{-1}$.
+
+$$
+\frac{d(1 / x)}{d x}= ? ? ?
```
3b1b, line 231: This is actually one we should explain.  Let's frame it as a multiple-choice question.

In the explanation, mention that the new area introduced by increasing the width is dx * (1 / x), and that the change in area caused by the reduction from the top is -x * d(1 / x).  Rearrange to find that d(1 / x) = (-1 / x^2) * dx.

Also mention that the little corner of area d(1 / x) * dx in the upper right has negligible influence as $dx \to 0$

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+$$
+
+
+## Comprehensive Questions
```
3b1b, line 235:
```suggestion
## Exercises
```

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+
+Here are a couple questions to test your knowledge of derivatives and the past chapters.
+
+### Derivative of $f(x) = \sqrt{x}$
+
+Modeling the derivatives geometrically can be applied to many different types of functions. For example, see if you can reason your way through the derivative of $f(x) = \sqrt{x}$ which also can be written as $x^{\frac{1}{2}}$. Now, one way to find the derivative of this function is to recognize that it fits the mold of the power rule. Blindly applying the rule will produce an answer, but what does it *mean*?
```
3b1b, line 241:
```suggestion
See if you can reason your way through the derivative of $f(x) = \sqrt{x}$ which also can be written as $x^{\frac{1}{2}}$. By far the easiest way to compute this is to apply the power rule: The exponent hops down as a coefficient, leaving behind $\frac{1}{2} - 1 = -\frac{1}{2}$ in the exponent.

$$
\frac{df}{dx} = \frac{1}{2}x^{-1 / 2}
$$

But is this valid? And following our current playful and geometric spirit, is there a way to read what this really *means*?
```

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+
+To start, try modeling function using geometry.
```
3b1b, line 243:
```suggestion
Approaching this question with geometry is, by far, overkill. 
Frankly, it's a bit of a mind warp. 
But it does offer a satisfying explanation of an otherwise mostly symbolic fact, and more than that it's one more opportunity to flex our muscles in reasoning about how small nudges to one value can affect another.
```

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+
+<Figure
+  image="figures/12.34.svg"
+  caption="Unlike in previous problems, $dx$ does not represent geometric length, but instead represents **area**."
+/>
+
+<Question
+  question="What is the derivative of the function $f(x) = \sqrt{x}$?"
```
3b1b, line 251:
```suggestion
  question="Using the diagram above, what is the derivative of the function $f(x) = \sqrt{x}$?"
```

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+  choice1="$\sqrt{2}x$"
+  choice2="$\frac{1}{2\sqrt{x}}$"
+  choice3="$\frac{x}{\sqrt{2}}$"
+  choice4="$\frac{1}{2}x$"
+  answer={2}
+>
+
+In order to find the derivative of the function $f(x) = \sqrt{x}$ we can model its geometry as the area of the square, where given the area $x$ the function returns the length of the square's side which is the square root of $x$.
```
3b1b, line 259:
```suggestion
Previously, to reason about $f(x) = x^2$, we thought of the input of the function as the side length of a square with the output being that square's area.
Here, for the function $f(x) = \sqrt{x}$, we do the opposite. We think of the input $x$ as being the area of a square, with the output $\sqrt{x}$ being the side length of that square.
```

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+$$
+
+Then, we want to find $d\sqrt{x}$, the change of the function's output, when the input $x$ is nudged by a small value $dx$. Since $x$ represents area, it makes sense that a small nudge $dx$ should also be area! 
+
+<Figure
+  image="figures/12.34-explanation.svg"
+  caption="Unlike in previous problems, $dx$ is not geometric length, but instead represents **area**."
+/>
+
+The added area changes the output of the function by the new length $d\sqrt{x}$, which looks like this.
```
3b1b, line 272: So the question is, how much does the side length of that square increase if the total change to the area is meant to be $dx$?

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+x + dx = x + 2\sqrt{x}(d\sqrt{x}) + (d\sqrt{x})^2
+$$
+
+Since we are only interested in the change in area, we can subtract $x$ from both sides. Additionally, we can safely ignore the expression $(d\sqrt{x})^2$ since the area it represents changes the output (length) of the function in a negligible way.
+
+$$
+dx = 2\sqrt{x}(d\sqrt{x})
+$$
+
+Finally, we can divide both sides by $dx$ and $2\sqrt{x}$ to get the expression below. The derivative of $f(x) = \sqrt{x}$!
```
3b1b, line 290:
```suggestion
Finally, we can divide both sides by $dx$ and $2\sqrt{x}$ to get the expression below.
```

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
```diff
@@ -1,10 +1,381 @@
...
+
+### Volume of Sphere
+
+In the first chapter, we found a formula for the area of a circle using a small nudge to the radius of the circle. Importantly, we derived the formula by finding the area under a function. In future chapters, we will explore this idea much further and derivatives have a big role to play. For now, let's revisit the same idea using the geometry of a sphere.
+
+<Figure
+  image="figures/sphere.svg" 
+/>
+
+<Question
```
3b1b, line 312: We should actually move this whole question much later to the chapter on integrals. It's not as relevant here, I think the sqrt question would be a fine place to end it.

3b1b (commented):
Okay, I reviewed everything through the power rule section.  Feel free to pull that out into its own article, and in the meantime, I'll try to hit the trig section when I get a chance.

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
kurtbruns: I agree

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
kurtbruns: I added an animation to the power rule figure.

public/content/lessons/2017/derivative-formulas-geometrically/index.mdx
kurtbruns: Sounds great, I had a feeling it was a bit out of place.

public/content/lessons/2017/derivatives-trig-functions/patrons.txt
```diff
@@ -0,0 +1,63 @@
...
+dim85
+Chris
+David Wyrick
+Rahul Suresh
+Lee Burnette
+John C. Vesey
+Patrik Agné
+Alvin Khaled
+ScienceVR
+Chris Willis
```
3b1b, line 36:
```suggestion
Grant Sanderson
```

public/content/lessons/2017/derivatives-trig-functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+timestamp: 756
+source: _2017/eoc/chapter3.py
+credits:
+- Lesson by Grant Sanderson
+- Text adaptation by Kurt Bruns
+---
+
+Let’s tackle one more type of function: Trigonometric functions. In particular, let’s focus on the sine function. Right here I’m going to assume you are familiar with how to think about trigonometric functions using the unit circle; the circle with radius $1$ centered at the origin.
+
+For a given value of theta, like $0.8$, you’d imagine walking around the circle, starting from the rightmost point, until you’ve traversed that distance $0.8$ in arc length. This is the same thing as saying that this angle here is exactly $0.8$ radians, since the circle has radius $1$. Then $\sin(\theta)$ is your height above the $x$-axis at this point.
```
3b1b, line 16:
```suggestion
For example, how would you interpret the value $\sin(0.8)$ if the value $\theta = 0.8$ is understood to be in radians?
You might imagine walking around a circle with a radius of $1$, starting from the rightmost point, until you’ve traversed the distance $0.8$ in arc length.
This is the same thing as saying you've traversed an _angle_ of $0.8$ radians. Then $\sin(\theta)$ is your height above the $x$-axis at this point.
```

3b1b: When the script includes phrases like "this angle here", that's usually with the intent that it's spoken while a specific thing is referenced on screen.  For the articles, sentences including those kinds of phrases are worth rewriting.

public/content/lessons/2017/derivatives-trig-functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+<Figure
+  image="figures/13.16.svg"
```
3b1b, line 19: Add a small arc and a label theta to emphasize that theta refers to the angle between the red line and the x-axis.

public/content/lessons/2017/derivatives-trig-functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+  video="figures/12.51-13.16.mp4"
+/>
+
+As theta increases, and you walk around the circle, your height bobs up and down and up and down, so the graph of $\sin(\theta)$ vs. $\theta$ is this wave pattern. The *quintessential* wave pattern.
```
3b1b, line 23:
```suggestion
As theta increases, and you walk around the circle, your height bobs up and down and up and down.
So the graph of $\sin(\theta)$ vs. $\theta$, which plots this height as a function of arc length, is a wave pattern. This is the *quintessential* wave pattern.
```

public/content/lessons/2017/derivatives-trig-functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+<Figure
+  image="figures/13.44.svg" 
+  video="figures/13.22-13.38.mp4"
+/>
+
+Now, just from looking at this graph, we can get a feel from the shape of the derivative function. The slope at $0$ is something positive, then as $\sin(\theta)$ approaches its peak, the slope goes down to $0$. Then the slope is negative for a little while before coming back up to $0$ as the $\sin(\theta)$ graph levels out. And, if you’re familiar with the graphs of trig functions, you might guess that this derivative graph should be exactly $\cos(\theta)$, whose graph is just a shifted-back copy of the sine graph.
```
3b1b, line 30:
```suggestion
Just from looking at this graph, we can get a feel from the shape of the derivative function. The slope at $0$ is something positive, then as $\sin(\theta)$ approaches its peak, the slope goes down to $0$. Then the slope is negative for a little while before coming back up to $0$ as the $\sin(\theta)$ graph levels out. If you’re familiar with the graphs of trig functions, you might guess that this derivative graph should be exactly $\cos(\theta)$, whose graph is just a shifted-back copy of the sine graph.
```

public/content/lessons/2017/derivatives-trig-functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+<Figure
+  image="figures/13.44-cosine.svg"
+  video="figures/13.40-14.30.mp4"
+  show="video"
+/>
+
+But all this tells us is that the peaks and valleys of the derivative graph seem to line up with the graph of cosine. How could we know that this derivative actually *is* the cosine of theta, and not just some new function that happens looks similar to it. Well, as with the previous examples of this video, a more exact understanding of the derivative requires looking at what the function itself represents, rather than just at the graph of the function.
```
3b1b, line 38:
```suggestion
But all this tells us is that the peaks and valleys of the derivative graph seem to line up with the graph of cosine. How could we know that this derivative actually *is* the cosine of theta, and not just some new function that happens looks similar to it. As with the previous examples of this video, a more exact understanding of the derivative requires looking at what the function itself represents, rather than the graph of the function.
```

public/content/lessons/2017/derivatives-trig-functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+So think back to the walk around the unit circle, having traversed an arc length of theta, where $\sin(\theta)$ is the height of this point. Consider a slight nudge of $d$-theta along the circumference of the circle; a tiny step in your walk around the unit circle. How much does this change $\sin(\theta)$? How much does that step change your *height* above the x-axis? This is best observered by zooming in on the point where you are on the circle.
```
3b1b, line 40:
```suggestion
Think back to the walk around the unit circle, having traversed an arc length of $\theta$, where $\sin(\theta)$ is the height of this point. Consider a slight nudge of $d$-theta along the circumference of the circle; a tiny step in your walk around the unit circle. How much does this change $\sin(\theta)$? How much does that step change your *height* above the x-axis? This is best observed by zooming in on the point where you are on the circle.
```

public/content/lessons/2017/derivatives-trig-functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+<Figure
+  image="figures/15.57.svg" 
+/>
+
+Well, zoomed in close enough the circle basically looks like a straight line in this neighborhood, so let’s go ahead and think of this right here as a right triangle, where the hypotenuse represents that nudge of $d \theta$ along the circumference, and this left side represents the change in height; the resulting tiny nudge to $\sin(\theta)$.
```
3b1b, line 46:
```suggestion
Zoomed in close enough the circle basically looks like a straight line in this neighborhood.
Consider the right triangle pictured below, where the hypotenuse represents a straight-line approximation of the nudge $d \theta$ along the circumference, and this left side represents the change in height; the resulting tiny nudge to $\sin(\theta)$.
```

public/content/lessons/2017/derivatives-trig-functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+<Figure
+  image="figures/16.00.svg" 
+/>
+
+Now this tiny triangle is actually similar to this larger triangle with the defining angle theta, and whose hypotenuse is the radius of the circle with length $1$. Specifically, this little angle is precisely equal to $\theta$ radians.
```
3b1b, line 52:
```suggestion
This tiny triangle is actually similar to this larger triangle with the defining angle theta, and whose hypotenuse is the radius of the circle with length $1$.
Specifically, the angle between its height $d(\sin(\theta))$ and its hypotenuse $d\theta$ is precisely equals to $\theta$.
```

3b1b: Here it would be helpful to show a side-by-side of the two relevant triangles. We should also include an explanation (perhaps presented as a challenge question). It basically comes down to the tiniest bit of angle chasing.

image

public/content/lessons/2017/derivatives-trig-functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+Think about what the derivative of sine is supposed to mean. It’s the ratio between that $d\left(\sin(\theta)\right)$, the tiny change to the output of sine, divided by $d \theta$, the tiny change to the input of the function. From the picture, that’s the ratio between the length of the side adjacent to this little right-triangle divided by the hypotenuse. Well, let’s see, adjacent divided by hypotenuse; that’s exactly what $\cos(\theta)$ means!
+
+<Figure
+  image="figures/16.01.svg" 
+/>
+
+So that gives two ways to think about how the derivative of $\sin(\theta)$ is $\cos(\theta)$. One by looking at the graph of sine and getting a loose feel for the loose shape of its derivative should be based on thinking about the slope of the graph, and another more precise line of reasoning that comes from looking straight at the unit circle. For the pause and ponder-ers among you, see if you can try a similar line or reasoning to understand the derivative of $\cos(\theta)$.
```
3b1b, line 60:
```suggestion
Notice, by considering the slope of the graph, we can get a quick intuitive feel for the rough shape that the derivative of $\sin(\theta)$ should have, which is enough to make an educated guess. But to more to understand why this derivative is *precisely* $\cos(\theta)$, we had to begin our line of reasoning with the defining features of $\sin(\theta)$.

For those of you who enjoy pausing and pondering, take a moment to find a similar line of reasoning which explains what the derivative of $\cos(\theta)$ should be.
```

3b1b: We should insert a diagrammatic explanation for the derivative of cosine under an accordion here.

It would be very similar to figures/16.01.svg, but the change $d(\cos(\theta)$ would be the width of the little rectangle. Again, there is a question of thinking carefully about the sign of this value. Because the x-coordinate of our little point is *decreasing* with a small step, that value $d(\cos(\theta)$ will be a negative number. Therefore, the width of the triangle will be $-d(\cos(\theta)$.

public/content/lessons/2017/derivatives-trig-functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+In the next video I’ll talk about how to take the derivatives of functions which combine simple functions like these, either as sums, products, functions compositions, things like that. Similar to this video, we’ll try to understand each rule geometrically, in a way that makes it intuitively reasonable and memorable.
```
3b1b, line 62:
```suggestion
In the <LessonLink id="chain-rule-and-product-rule">next lesson</LessonLink> we'll figure out the derivatives of functions that combine simple functions like these, either as sums, products, or functions compositions. Similar to this lesson, we’ll try to understand each rule geometrically, in a way that makes it intuitively reasonable and memorable.
```

PR #145, opened 2021-06-28
Homepage design

### LINKED ISSUE #140: Home page design
After our discussion, I do agree that a slightly different home page landing experience could be better and that pulling out the featured lessons to be separate from the lesson gallery makes sense. It seems like there are potentially four things to include at the top of the home page.

- Brief words about what the site/channel is.
    - Possibility: "3blue1brown videos aim to illustrate the sense of creativity and discovery in math, presented with a highly visual style."
- Latest video
- Featured content
    - Maybe as a shelf beneath the top spotlight?
- Lesson gallery

The big benefit of the current single-video-in-the-spotlight vibe is its minimalism. This is potentially a nice way to set the tone of the minimalism of the content itself, which often aims to just display what is needed and nothing more. Even if we want to include all four things above, it would be best to do so in a way that still sets this tone.

@kurtbruns, if you have a chance, would you like to quickly throw together a wireframe or two to work from? I think the answer could be pretty simple, maybe just the words and latest video presented side-by-side up top, with a shelf of ~4 featured lessons below, followed by the lesson gallery. But sometimes the devil's in the details.

PullJosh:

This pull request:
- Adds a new homepage design with a rotating carousel of featured lessons
- Changes the way that interactives are sized on the page. Interactive authors can now write their code as if the interactives are a fixed size (such as, for example, 880x500 pixels), and then the `<Interactive>` component will automatically scale the contents similarly to how images and videos are scaled to fit within the page. (This was necessary to make the neural network interactive work both in the article and on the homepage.)
- Closes #140


3b1b:
I know this is only partially finished, so perhaps too early for feedback, but here are some nates as I look at it:
- Looking at it now, I might abbreviate the text simply to "Lessons about creativity and discovery in math."
- I'd have not autoscrolling on the featured carrosel.
- For cases where the featured item does not match the 16x9 aspect ratio, such as the neural nets one, I'd match the height instead of the width.
- Rather than cycling around, I'd say don't display the left arrow for the first item in the list and don't display the right arrow for the last item in the list.

PullJosh:
Now's a perfect time for feedback. And I think every point you made is smart. I will implement it tomorrow.

PullJosh:
@3b1b Okay, I've implemented your suggestions: https://deploy-preview-145--3blue1brown.netlify.app/

Thoughts now?

3b1b:
Looks great, thanks!  Only two notes things before merging.

- Instead of darkening the left arrow when it's inactive, I'd remove it entirely.
- For some reason, the animations don't play on mobile for me, on Chrome/Safari/Brave. I imagine this could be solved just by making them gifs, but gifs tend to be some combination of lower quality and more data.

PullJosh:
> Instead of darkening the left arrow when it's inactive, I'd remove it entirely.

I actually like the symmetry of having both arrows present. It helps indicate that scrolling is an option, you've just reached the end of the slides.

> For some reason, the animations don't play on mobile for me, on Chrome/Safari/Brave. I imagine this could be solved just by making them gifs, but gifs tend to be some combination of lower quality and more data.

Oh, weird. Me too. I'll try to figure it out.

3b1b:
> I actually like the symmetry of having both arrows present...

Works for me, it's a very soft opinion.

PullJosh:
I'm pretty sure the video issues are fixed on mobile. It works on my phone now (and it didn't before).

3b1b:
> I'm pretty sure the video issues are fixed on mobile. It works on my phone now (and it didn't before).

Yup, works on mine too. I'm happy with merging that if you are.

PR #147, opened 2021-06-29
Use a thumbnail cover for unplayed videos

### LINKED ISSUE #142: YouTube videos embedding at the top of lesson pages.
- At the moment, before playing the YouTube player, it shows the title and other info at the top which clutters things and is redundant with the title listed below. Moreover, when chapter numbers deviate it might actively conflict.

    Screen Shot 2021-06-27 at 2 34 05 PM

    It looks like YouTube deprecated the embed parameter which hides that info, but perhaps we can simply show the thumbnail with a YouTube-style play button in the middle, like [this](https://kurzgesagt.org/portfolio/the-bacteriophage/).

- While I think the experience landing on the page to look around is nicer/more balanced if the embedded video is 70% width, the reason I opted for 100% is that once you're actually watching the video I believe it's nicer at 100%. I'm also not a fan of video embeds that force someone to click full screen to have a reasonable experience ([ahem](https://www.minutephysics.com/)).  If there's a natural and non-janky way to have the embed start at 70%, then expand upon being played, it may be worth exploring.

PullJosh:

This pull request:
- Adds a thumbnail cover to videos that is smaller than the full video, so that the video can be small when unplayed but larger when played
- Closes #142


3b1b:
Great, on the whole, this looks good, and I think it'll be the way to go.

- I would make the play button smaller, changing the width from 15% to 8% feels about right to me.
- It would be nice to mimic the behavior of the YouTube player where that play button is translucent black until the mouse hovers over the video when it turns red.
- Upon playing, I might even expand the video as big as is reasonable, filling more of the available height/width of the screen.

3b1b:
Although, if https://github.com/3b1b/3Blue1Brown.com/issues/148 seems ever seems like a way to go, then the video shouldn't expand to be _too_ big.

PullJosh:
I really like #148, so let's hold off on making the video any bigger for now. We can always adjust it later.

PR #150, opened 2021-06-30
Lesson navigation

### LINKED ISSUE #148: Replace top-of-page lesson navigation
The articles may look cleaner without the navigation at the top. It's a feature I initially requested, but something feels off about the very first part of an article being an invitation for where else to go. It also breaks the visual flow a bit.

image

Since we switched to having topic pages, perhaps another way to accomplish the same goal of easy navigation when landing on a page is to move the next/previous/topic functionality to buttons around the video itself, say like this.

image

PullJosh:

This pull request:
- Changes the design for navigation at the top of lesson pages
- Closes #148


3b1b:
On mobile I don’t see any next/prev arrows, is that right?  They could probably sit below the videos as you did for the featured graphics.

PullJosh:
Yeah. I figured hiding them would be cleanest, but below definitely works too. 👍

PullJosh:
@3b1b Thoughts now?

3b1b:
I like it!  Just a couple of notes

- Upon clicking, either the page should scroll down just a bit, to the section holding the video, or the video should expand slightly less.  At the moment, after clicking the video falls off the bottom of my screen ever so slightly.
- After starting to play one video, then clicking the next or previous arrow, the video on the following page should start in its initial small not-playing state.  It seems like the state of "video is large and playing" gets automatically carried over at the moment.

3b1b:
That looks great, thanks!

PR #168, opened 2021-07-02
Add free response component

### LINKED ISSUE #159: Free response component
Have a way to ask questions which invites users to type their own answer. It would probably be something super simple, maybe like [this](https://itempool.com/3b1b/c/yfTpkVn5WpN)?

Upon clicking "check", the user would see an example answer, perhaps labeled as "Possible answer"

It may be worth having a tiny explanation of the widget somewhere, say a small question mark in the corner of the component with a tooltip, where we say explicitly that this is not a checked or saved answer, it's just a prompt for readers to write down and clarify their own thinking.

PullJosh:

This pull request:
- Adds a free response component
- Includes a free response question in the testbed
- Closes #159

### DIFF
--- a/components/FreeResponse/index.js
+++ b/components/FreeResponse/index.js
@@ -1,7 +1,6 @@
 import { useState } from "react";
 import Clickable from "../Clickable";
 import Tooltip from "../Tooltip";
-import TextareaAutosize from "react-textarea-autosize";
 import styles from "./index.module.scss";
 
 export default function FreeResponse({ children: explanation }) {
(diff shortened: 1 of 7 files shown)


3b1b:
This looks nice, and I love the tooltip text. Just two quick notes:

- I'd make the submit button smaller, and place it outside the textbox. I think hanging off the lower right might be fitting.
- Similarly, I'd put the question mark outside the box, say above the box aligned to the right.

The choice to lock the text upon submitting was an interesting one. My initial reaction was that it felt unnecessary, but looking at it more it seems to give a kind of significance to the submit button which would otherwise not really be there as much.

PullJosh:
Okay, I've changed the styles based on your recommendations. I wasn't confident that it would look good, but now that I see it this way I really like it. 🤩

> The choice to lock the text upon submitting was an interesting one. [...]

That's interesting! I didn't even give that feature a second thought. It just seemed necessary.

(I think the lesson here is that having more than one set of eyeballs on something really helps.)

PullJosh:
I just realized you can't see the testbed on the preview. So here's a screenshot for anyone who is too lazy to check out the branch:

image

3b1b:
Looks good to me, merge away.

PR #174, opened 2021-07-03
Content tweaks

3b1b:

- [ ] I have requested a review from the appropriate persons
- [ ] I have checked that the rest of the site is still functioning in the preview
- [ ] I have checked that my content/code is consistent with existing content/code
- [ ] I have used components in the places and ways they are intended (see the readme)
- [ ] I have formatted all of my code with Prettier

- Several small tweaks/fixes
- Add an expand/compress button to the LessonVideo


PullJosh:
Everything looks good, except that the wide video is *very* wide on my laptop screen:

image

You're setting a max width of 1500px, but my screen isn't that wide, so the video stretches right up to the very edges, which means that it just barely fits vertically.

3b1b:
> except that the wide video is very wide on my laptop screen

Noted.

> It also feels weird to reinvent the fullscreen button. Are we making things more complicated than they need to be?

You're asking the right question. I still think it makes sense to have a smaller thumbnail by default, so as to not dominate the article. I also think it makes sense to have it grow if someone clicks on it since I rather dislike embeddings that are too small.  But with both those, from a UX perspective, it felt weird not to be able to go back; so I thought I'd try throwing in a "compress" button just to see how it felt.

PR #176, opened 2021-07-07
Add topology interactive

PullJosh:

This pull request:
- Adds the interactive 3D surface to the topology article (`inscribed-rectangle-problem`)
- Adds a comprehension question (free response) about the interactive


PullJosh:
@3b1b Thoughts on the free response question I added?

3b1b:
I love the question, and how well it prompts someone to play more deliberately with the interactive.  If it was me, I might tweak the explanation to be a little more exact.  Say something like this:

"The pair of points at opposite ends of the straight segment produce a point in 3D space at the very top of the triangle. As one point moves inward on the segment, the distance $d$ between the points decreases, and the midpoint remains on the line segment a distance $d/2$ from each of them, so the resulting point in 3D space traces out a downward sloping line, specifically with slope 2, which forms the edge of the triangle. Moving both points inward produces points that fill in the area under the triangle."

PR #186, opened 2021-07-08
Visual Fourier Article (Next edition)

riverliway:

- [x] Includes authorship line: `Text adaptation by [Name]`
- [x] The lesson has an informative/compelling description, fit for a preview on the home page.
- [x] The lesson has no grammar/spelling errors
- [x] The lesson has no mathematical errors
- [x] The table of contents effectively conveys the overall structure of the lesson
- [x] The lesson uses no `# Top Level` headings
- [x] The lesson includes well-chosen comprehension questions
- [x] The lesson incorporates the core visuals from the video
- [x] The visuals are clear and helpful for understanding
- [x] The components are used in the places and ways they are intended (see the readme)


public/content/lessons/2018/fourier-transforms/index.mdx
```diff
@@ -6,4 +6,343 @@ video: spUNpyF58BY
...
+  image="almost.png" width="100%"
+/>
+
+The way a mathematician might phrase this is that our "Almost Fourier transform" operation is *linear*. The curious among you may want to pause for a moment to reflect on why this is true. It ultimately stems from the fact that the x-coordinate of the sum of two vectors is the sum of their x-coordinates. This, by the way, is why we had it track the x-coordinate of the center of mass of the graph, rather than tracking its distance from the origin. For the full frequency information, you'd also want to keep track of the y-coordinate, but more on that below.
+
+So this little mathematical machine does exactly what we wanted: it pulls out the constituent frequencies from a provided waveform, essentially unmixing a mixed bucket of paint. The applet below allows you to change the input wave form by separating the frequencies by commas.
+
+<Interactive filename="advanced_interactive/index" aspectRatio={880 / 500} />
+
+What happens when an input frequency of $0$Hz is wound up?
```
kurtbruns, line 186: Is the dotted line supposed to represent a neutral pressure? Because if that's the case, should 0Hz be constant at that line?

kurtbruns (commented):
The interactive look great! I left a comment about possibly making the scrubbing a bit more snappy.

public/content/lessons/2018/fourier-transforms/index.mdx
```diff
@@ -6,4 +6,343 @@ video: spUNpyF58BY
...
+
+<Figure
+  image="com_1.png"
+  video="com_1.mp4"
+  width="100%"
+/>
+
+Different patterns appear as we wind up this graph, but it is clear that the $x$-coordinate for the center of mass is important when the winding frequency is $3$. Check out how the red dot marking the center of mass moves as the winding frequency changes. The interactive applet below will respond to hovering your mouse over the plot of the $x$-coordinate.
+
+<Interactive filename="basic_interactive/index" aspectRatio={880 / 500} />
```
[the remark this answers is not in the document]
riverliway: I think the speed of the scrubbing also is related to the slow animations I discovered while watching Grant's screen. I realized now that the draw function is just called instantly after the previous draw has completed. I thought it was called at some fixed framerate, but it executes much faster on my computer than it did on Grant's. So I'll change the code so that the scrubbing is time based instead of number of draw frames. That way the scrubbing will be the same time for every device. Then we can pick a momentum value that is more snappy.

public/content/lessons/2018/fourier-transforms/index.mdx
riverliway: Yes the dotted line is supposed to represent neutral pressure, great catch! I decided to normalize the resulting wave because if you had a wave composed of 1+2+3+4 then it would be drawn partially off the screen because the amplitude of that wave is simply larger than a wave composed of a single frequency. But 0 Hz should not affect the normalized wave!

riverliway:
Todos:

[ ] Time based scrubbing, not frame count based
[ ] Less momentum on scrubbing, make more snappy
[ ] Better communication to user for frequency limitations (either switch to sliders or make a tooltip)
[ ] 2nd applet needs to fix aspect ratio

public/content/lessons/2018/fourier-transforms/index.mdx
riverliway: I've thought about it some more and I actually am thinking that it is actually fine the way it is. A signal with 0Hz is just a constant line, which could mean that it is always at the peak instead of always at neutral pressure.

PR #203, opened 2021-07-21
Improve search result order of lessons in a series

### LINKED ISSUE #189: Order of items in search
While something like https://github.com/3b1b/3Blue1Brown.com/issues/164 might be a bit out of scope at the moment, there may be simpler lower hanging fruit to make the order of search results a bit better.

The main niggle I have at the moment is that if you search a topic, like linear algebra, that topic shows up in reverse order.

image

A very simple way around this could be to use the order of topics as they show up in data/topics.yaml

PullJosh:

This pull request:
- Slightly tweaks the search result ordering so that most results are reverse-chronological, but lessons in a series try their best to appear in a reasonable order, with probably an 85% success rate
- Makes neural networks a series with chapter numbers (rather than just a topic)
- Closes #189

### DIFF
--- a/components/LessonGallery/index.js
+++ b/components/LessonGallery/index.js
@@ -28,7 +28,21 @@ export default function LessonGallery({ show = "topic" }) {
       return lessons;
     }
 
-    return lessons.filter((lesson) => matchesSearch(lesson, searchText));
+    return lessons
+      .filter((lesson) => matchesSearch(lesson, searchText))
+      .sort((a, b) => {
+        if (a.topic === b.topic) {
+          if (typeof a.chapter === "number") {
+            const topic = topics.find((t) => t.name === a.topic);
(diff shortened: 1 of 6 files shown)


PullJosh:
@3b1b I'm experimenting with a little tweak to the sorting algorithm that makes it *mostly* reverse-chronological, but with some exceptions for lessons in a series. It's not perfect, but it feels like the kind of solution where 20% of the effort gives you 80% of the benefit.

If it feels good to you, I'll merge.

3b1b:
Seems to work pretty well. I'm trying to understand why calculus chapter 12 shows up at the top of a "calculus" search, but honestly, it doesn't really matter.

> It's not perfect, but it feels like the kind of solution where 20% of the effort gives you 80% of the benefit.

I agree! Looks good to me, merge away.

PR #234, opened 2021-07-29
Calculus limits, epsilon delta and l'hopital

kurtbruns:
[Netlify PR Preview Limits](https://deploy-preview-234--3blue1brown.netlify.app/lessons/limits)
[Netlify PR Preview Epsilon Delta](https://deploy-preview-234--3blue1brown.netlify.app/lessons/epsilon-delta)
[Netlify PR Preview L'Hopital](https://deploy-preview-234--3blue1brown.netlify.app/lessons/l-hopitals-rule)

This pull request:

- Publishes three calculus lessons: limits, epsilon delta and l'hopital's rule
- Modifies compress command to remove audio and output at 1080p quality


3b1b:
Looking really good! As usual the imagery is quite well chosen, and well-designed.

Notes on the definition of derivative chapter:

- Can we add some kind of comprehension question to the definition of the derivative?
- (This one’s on me) the small rant could be trimmed down even more. Just a quick emphasis of the main point, which in the series has maybe even been belabored as it is.
- I might rewrite the last paragraph as follows:

> Even more than this, mathematicians care about ensuring all their statements have a precise and airtight meaning, which does not rest on visual intuition. As sensible as it might seem to say that this ratio approaches a certain value, and to draw a line through the graph whose slope approaches that of a tangent line, those seeking rigor insist one something better. In the next chapter, we’ll take a look the <LessonLink>epsilon-delta</LessonLink> definition of limits, which finally puts this idea of “approaching” onto a firm foundation.

3b1b:
I know it’s not finished, but I thought I’d take a look and leave some ambient notes as I did.

Notes on the epsilon-delta chapter:
- I’m only just noticing this, but none of the animations seem to be working (at least from my iPad)
- Arguably, we could delete the paragraph starting with “ You could argue this needlessly heavy-duty for an introduction to calculus...”
- I would eliminate the use of the words “for some reason”
- Good comprehension questions.

Notes on L’Hopital:
- So, in this lesson, let’s finish off our dive into limits [with a?] trick for actually computing limits.
- dampened -> damped
- “it looks pretty continuous” -> “it looks continuous”
- Delete “That’s kind of a lot to look at, but”
- For an explanation to the first comprehension question, when you do fill it in I’d try to keep it short and sweet. Highlight how when you evaluate the original function at f(1 + dx), what you get is this ratio d(sin(pi*x)) / d(x^2 + 1).
- Maybe it’s worth somewhere highlight that we’re basically taking the ratio of the slope of the yellow line to the slope of the blue?
- In the image above the phrase “ Remember, what that means is that the limiting”, I’d include an intermediate equality to the expression $\lim_{x \to 0} \frac{\pi \cos(\pi x)}{2\pi}$

kurtbruns:
@3b1b Thanks for jumping on this so I was able to start on it straight away this morning.

I made most the edits that you suggested. If you have time, you could take a look at the word-smithing I did to trim the rant size down and the draft of L'Hopital's rule that I finished today.

I am still thinking of a good comprehension question for the definition of the derivative. Maybe one will come to me during lunch.

Looking like these lessons can be merged by end of day!

PR #346, opened 2023-05-25
Essence of calculus: Taylor series

kurtbruns:
Opening this pull request for the first stage review of the Taylor Series article.

Notes:

Does it make sense to separate out the geometric view of taylor series into a separate article? In the middle, the article feels a bit wordy and I'm worried about a reader getting lost in the sauce.


public/content/lessons/2017/taylor-series/index.mdx
```diff
@@ -7,4 +7,376 @@ video: 3d6DsjIBzJ4
...
+/>
+
+This is true for any input, no matter how far away from $0$ it is, even though these Taylor polynomials are constructed only from derivative information gathered at the input $0$.
+In a case like this, we say $e^x$ equals its Taylor series at all inputs $x$, which is kind of a magical thing to have happen.
+
+## Limitations for $\ln(x)$
+
+Although this is also true for some other important functions, like sine and cosine, sometimes these series only converge within a certain range around the input whose derivative information you're using.
+
+If you work out the Taylor series for the natural log of $x$ around the input $x = 1$, which is built from evaluating the higher order derivatives of ln(x) at x=1, this is what it looks like.
```
3b1b, line 348:
```suggestion
If you work out the Taylor series for $\ln(x)$ around the input $x = 1$, which is built from evaluating the higher order derivatives of $ln(x)$ at $x=1$, this is what it looks like.
```

public/content/lessons/2017/taylor-series/index.mdx
```diff
@@ -7,4 +7,376 @@ video: 3d6DsjIBzJ4
...
+
+When you plug in an input between $0$ and $2$, adding more and more terms of this series will indeed get you closer and closer to the natural log of that input.
+
+<Figure
+    image="figures/natural-logarithm-taylor-series-input-between-0-and-2.png" 
+/>
+
+But outside that range, even by just a bit, the series fails to approach anything.
+
+As you add more and more terms the sum bounces back and forth wildly, it does not approaching the natural log of that value, even though the natural log of $x$ is perfectly well defined for inputs above $2$.
```
3b1b, line 362:
```suggestion
As you add more and more terms the sum bounces back and forth wildly.  The partial sums do not approach the natural log of that value, even though the $ln(x)$ is perfectly well defined for $x > 2$.
```

public/content/lessons/2017/taylor-series/index.mdx
```diff
@@ -7,4 +7,376 @@ video: 3d6DsjIBzJ4
...
+
+<Figure
+    image="figures/natural-logarithm-taylor-series-input-above-2.png" 
+/>
+
+In some sense, the derivative information of $\ln(x)$ at $x=1$ doesn't propagate out that far. In a case like this, where adding more terms of the series doesn't approach anything, you say the series **diverges**.
+
+And that maximum distance between the input you're approximating near, and points where the outputs of these polynomials actually do converge, is called the "radius of convergence" for the Taylor series.
```
3b1b, line 370:
```suggestion
That maximum distance between the input you're approximating near, and points where the outputs of these polynomials actually do converge, is called the "radius of convergence" for the Taylor series.
```

public/content/lessons/2017/taylor-series/index.mdx
```diff
@@ -7,4 +7,376 @@ video: 3d6DsjIBzJ4
...
+    image="figures/natural-logarithm-radius-of-convergence.png" 
+/>
+
+## Summary
+
+There remains more to learn about Taylor series, their many use cases, tactics for placing bounds on the error of these approximations, tests for understanding when these series do and don't converge. For that matter there remains more to learn about calculus as a whole, and the countless topics not touched by this series.
+
+The goal with these videos is to give you the fundamental intuitions that make you feel confident and efficient learning more on your own, and potentially even rediscovering more of the topic for yourself. In the case of Taylor series, the fundamental intuition to keep in mind as you explore more is that they translate derivative information at a single point to approximation information around that point.
+
+TODO: maybe add "and in the case of the exponential function where the series converges for all values, this can be used as a definition of the exponential function throughout other areas of math."
```
3b1b, line 382: Maybe!  If there was ever an article on the matrix exponents video, we could also just link out to that.

3b1b (commented):
Thanks for the draft! I've added many inline suggestions, mainly based on ways that I would change what I wrote for the video script back then to fit better in written form.

Almost all of the images feel like well-chosen starting points for any custom visuals here. One thing I'd say is that for screenshots capturing a part of the video where the primary function of the video was to show something symbolic, like a formula, we should just lean on things like inline Tex expressions. A single image of the relevant graph at the top of a section can set the visual context well enough that it doesn't necessarily have to be intertwined with the formulas the way it showed up in the videos.

Henrisen (approved):

PR #352, opened 2023-07-06
Update index.mdx

aartvanbochove:
Added a comment to show an error that should be fixed

Text adaptation by Aart van Bochove-Smid


3b1b:
Thanks for pointing it out! I've fixed it on another branch now.

PR #368, opened 2023-12-18
Display Images In Production

### LINKED ISSUE #367: Images Not Displaying
**Describe the bug**
Images are available at `https://www.3blue1brown.com/images/**/*`, but do not show up on various pages such as `https://www.3blue1brown.com/about`

**To Reproduce**
1. Visit https://www.3blue1brown.com/about

**Expected behavior**
image
image

shahdivyank:

- Fixes #367 
- adds an images field in the config with an empty path to allow for relative images to pull from `BASE_URL/images/**/*` instead of `/images/**/*`
- Cannot be verified in dev, since dev continues to showcase the same behavior, needs to be pushed and built to check production
- I have implemented this change across some other frontend websites in the past and it has solved

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -18,6 +18,9 @@ config = {
 
 config = {
   ...config,
+  images: {
+    path: "",
+  },
   webpack(config) {
     config.module.rules.push({
       test: /\.svg$/,


3b1b:
Thanks!

PR #387, opened 2024-01-11
Publish Cross Products Extended Lesson

kurtbruns:
This pull request:

- Publishes the cross-products-extended lesson
- Publishes 3d illustrations for the cross-products lesson


public/content/lessons/2016/cross-products-extended/index.mdx
```diff
@@ -1,10 +1,256 @@
...
+    image="./figures/the-idea/DotProductBetweenPAndU.svg" 
+/>
+
+So the answers to what those "somethings" are will give the coordinates of the vector $\vec{\mathbf{p}}$ that we're looking for.
+
+<Figure
+    image="./figures/the-idea/DotProductBetweenPAndUComponentsOfP.svg" 
+/>
+
+But this should feel very familiar to anyone who's actually worked through a cross product computation before: Collecting the constant terms that are multiplied by $x$, $y$ and $z$ like this is no different from plugging in the symbols $\hat{\imath}$, $\hat{\jmath}$ and $\hat{k}$, and seeing which coefficients aggregate on each of these terms. It's just that plugging in $\hat{\imath}$, $\hat{\jmath}$ and $\hat{k}$ is a way of signaling that we should interpret these three coefficients as coordinates of a vector.
```
kurtbruns, line 132: Because of the line "But this should feel very familiar to anyone who's actually worked through a cross product computation before" I think it makes sense to add some plug-and-chug type practice problems to the regular cross products lesson. Is the best place for those questions at the very end in their own section?

public/content/lessons/2016/cross-products-extended/index.mdx
```diff
@@ -1,10 +1,256 @@
...
+
+For both approaches, finding the dual vector of this linear transformation gives us a much deeper understanding of the result of the cross product between two vectors $\vec{\mathbf{v}}$ and $\vec{\mathbf{w}}$ than just the formula.
+
+<Figure
+    image="./figures/conclusion/CrossProductFormula.svg" 
+/>
+
+Next up are two important concepts in linear algebra: <LessonLink id="cramers-rule">Cramer's Rule</LessonLink> and <LessonLink id="change-of-basis">Change of basis</LessonLink>.
+
+## Practice
```
kurtbruns, line 219: Does it make sense to try and get the student to think along the lines of both the computational and geometric approach using practice questions in this lesson? I struggled to think of a natural place to inject a question into “the idea” section, but perhaps questions at the end could serve to test the students comprehension of both the notational trick (by gathering like expressions) or finding the signed volume of a parallelepiped and then finding the dual vector associated with the transformation function that related to the parallelepiped. Or, alternatively, is it enough to provide some questions about the cross-product of two vectors, and leave it up to the student to try and apply one or both approaches?

public/content/lessons/2016/cross-products-extended/index.mdx
3b1b: No strong preference, but we could throw one such exercise at the end of the first Cross Product article.

public/content/lessons/2016/cross-products-extended/index.mdx
3b1b: That's a good question. One possible comprehension question would be to give the coordinates of three vectors, v, w, u, and ask for v dot (w cross u), and also (v cross w) dot u, and to mention in the explanation how if one understands the meaning here, you could find it by calculating the 3x3 determinant of a matrix holding all their vectors.

3b1b:
The graphics look great, and it adds a lot to clarify a potentially tricky idea. Even though the last exercise could still use an explanation, I'll go ahead and merge this for now.

PR #394, opened 2024-01-24
Add an initial component for gallery of past contributors

3b1b:


vincerubinetti:
Looks good, just made some minor tweaks and updated my headshot.

PR #396, opened 2024-01-26
Essence of Linear Algebra: Revised Span Lesson

kurtbruns:
This pull request publishes these changes for the "span" lesson:

- Publishes vectorized images consistent with the rest of the series
- Adds 3d visuals with questions and animations, replacing the existing interactives
- Changes author's perspective to plural
- Removes a section about inconsistent notation from the article


public/content/lessons/2016/span/index.mdx
```diff
@@ -7,120 +7,151 @@ video: k7RM-ot2NWY
...
+  image="./figures/different-basis-vectors/DifferentBasisSpanExample3.svg" 
+  caption="$\left[\begin{array}{c}-0.8 \\ 1.3\end{array}\right] \neq\left[\begin{array}{c}3.1 \\ -2.9\end{array}\right]$"
 />
 
-In the above figure, $\overrightarrow{\mathbf{v}}=\begin{bmatrix}1\\ 2\end{bmatrix}$ and $\overrightarrow{\mathbf{w}}=\begin{bmatrix}3\\ -1\end{bmatrix}$.
+We'll go into much more detail on this point in a <LessonLink id="change-of-basis">later chapter</LessonLink>, describing the relationship between different coordinate systems, but for now we just want you to appreciate that any way to describe vectors numerically depends on your choice of basis vectors.
 
 <Question
-  question="What values of the scalars $\alpha$ and $\beta$ satisfy the following equation? $\alpha\vec{\mathbf{v}}+\beta\vec{\mathbf{w}}=5\hat i -\frac12\hat j$"
+  question="Given the equation $\alpha \vec{\mathbf{v}} + \beta \vec{\mathbf{w}} =5\hat{\imath} -\frac12\hat{\jmath}$, where $\vec{\mathbf{v}} = \left[\begin{array}{c} 1 \\ 2 \end{array}\right]$ and $\vec{\mathbf{w}} = \left[\begin{array}{c} 3 \\ -1 \end{array}\right]$, what are the values of the scalars $\alpha$ and $\beta$?"
```
kurtbruns, line 79: I think there is a general question of when is the appropriate time to introduce solving a linear system of equations to the student. This is a question from River, see the new question at the end of this lesson as well as the question "formal properties" section in the revised "linear transformations" lesson. I think the answer may be at the end of this article, but maybe we should preface or introduce the idea.

3b1b:
Great, this is an improvement. I noticed that the previous interactives were simply not working on my iPad with Safari.

The only thing that caught my eye as worth changing is that for the question "What is the span of...", the target audience may not have seen the notation $\mathds{R}^2$ before. It also doesn't quite make sense to compare it with R and 0, since in all cases the span is a subset of R^2, i.e. a set of pairs of numbers, not a set of numbers or an individual number. The goal of the question could be better accomplished just with the options A) A point, B) A line C) The full 2d plane.

public/content/lessons/2016/span/index.mdx
kurtbruns: I removed this earlier question in favor of the later one.

PR #397, opened 2024-01-27
Essence of Linear Algebra: Revised Linear Transformations Lesson

kurtbruns:
This pull request publishes these changes for the "linear transformations" lesson:

- Publishes vectorized images and animations consistent with the series
- Adds multiple new questions to the article
- Changes perspective to be plural
- Adds attribution for James

Notes:

- Structurally the biggest change was moving the "formal properties" into their own section. I think this best preserves the original intent that the lesson be visually motivated. Note: these formal properties are explicitly referenced in the "dot products" lesson. Possibly they could live in a stand-alone lesson. I didn't realize this part was scripted as it's just a frame at the very end frame of the video, so I modified what River had already written. See the very end of the [original script](https://docs.google.com/document/d/14PWlqgvW5UMAA0LnOosxpYg5WFN2P4PVKde564VHLUo/edit) for reference.
- @3b1b What are your thoughts on keeping the original input vector in the visuals during the "transformations are functions" section?


3b1b:
Looking good. There are just a few small things I noticed:
-

public/content/lessons/2016/linear-transformations/index.mdx
```diff
@@ -42,355 +36,534 @@ So why use the word “transformation” instead of “function” if they mean
...
-+y\begin{bmatrix}\color{red}b\\ \color{red}d\end{bmatrix}
-=
-\begin{bmatrix}
-\color{green}a\color{black}x+\color{red}b\color{black}y \\
-\color{green}c\color{black}x+\color{red}d\color{black}y
-\end{bmatrix}
-$$
-Remember that this all came from thinking about the columns as the transformed versions of your basis vectors. Then the result is the appropriate linear combination of those vectors.
+<Figure
+  image="./figures/matrices/MatrixNotation2x2InputOutput.svg"
```
kurtbruns, line 243: I agree this illustration and the one below was a bit confusing. I just updated this illustration and the one below to something like this. What do you think?

Screenshot 2024-01-30 at 12 26 36 PM

3b1b: Much better, thanks!

Maybe there's not room to do it cleanly, but you could also try one where the arrow is extra long, and the matrix is small and sits on top of the arrow, as a way of communicating that it is the transform.

PR #399, opened 2024-01-29
Fix broken lesson contributor links

kurtbruns:
This pull request:

- Updates `credits.yaml` to contain valid links to contributors in the "about" page when navigating from a lesson
- Adds padding to the top of team member cards so there is some breathing room when navigating to a team contributor's info. For example, when navigating to: `https://www.3blue1brown.com/about#james-schloss`.
- Updates Kurt's social links

Notes:

The `credits.yaml` file could probably be removed entirely, but it would require modifying some logic in `components/LessonDetails/index.js`.


3b1b:
Thank you!

PR #403, opened 2024-02-09
Essence of Linear Algebra: Publish 3d transformations lesson

kurtbruns:
This pull request:

- Publishes the 3d transformations lesson for review

Notes for review:

- Do you prefer the standard 3d basis vectors to be illustrated with the y-direction pointed away?
- This is a bit pedantic, but should we be using the maps-to symbol (↦) instead of the right-arrow symbol (→) when referring to the specific input and output of a function?
- Is the explanation too verbose for the Combining Transformations explanation? Does a free-response question make sense here?  Or should we specifically prompt them with "Given the two transformations... what is the combined transformation?"
- Similarly, the puzzle feels a bit open-ended and is answered later in the "non-square matrices" lesson which was tacked on because of questions from the audience.


3b1b:
This pull request:

Publishes the 3d transformations lesson for review
Notes for review:

> Do you prefer the standard 3d basis vectors to be illustrated with the y-direction pointed away?

Often I do it that way, yes.

> This is a bit pedantic, but should we be using the maps-to symbol (↦) instead of the right-arrow symbol (→) when referring to the specific input and output of a function?

That seems slightly better, just as long as its consistent.

> Is the explanation too verbose for the Combining Transformations explanation? Does a free-response question make sense here? Or should we specifically prompt them with "Given the two transformations... what is the combined transformation?"

First, I would put it inside an accordion instead of the free response, perhaps prompting a reader to think about it. It is too verbose, though. To state it more simply, we need to follow where a given basis vector goes, for instance, i -> M1 * i -> M2 * M1 * i.  After the first step, its coordinates are the first column of M1. After the next step, its coordinates are M2 * (that first column).

> Similarly, the puzzle feels a bit open-ended and is answered later in the "non-square matrices" lesson which was tacked on because of questions from the audience.

3b1b:
As a side note, I think we should edit the free response question to not have a textbox since it's hard to envision anyone literally typing for anything more than an answer with a few characters. It should instead be labeled something like "Pause and ponder", with the same blurred-out result with a button to reveal.

PR #404, opened 2024-02-09
Essence of Linear Algebra: Publish nonsquare matrices lesson

kurtbruns:
This pull request:

- Publishes the nonsquare matrices lesson

Notes for review:

- Is it worthwhile challenging the viewer with the question about a stereographic projection at the end? If so, I'll continue and add an explanation.
- What do you think about a conceptual question about projecting a 3d scene to a viewport on a flat device? The question would be something like: "Consider the function that maps coordinates in this 3d scene to the 2d device that you are viewing it on. Is this function linear (yes/no)?
 
PerspectiveQuestion

Hint:

PerspectiveQuestionHint


3b1b:
> Is it worthwhile challenging the viewer with the question about a stereographic projection at the end? If so, I'll continue and add an explanation.

The way the question is phrased now, there's no reason it'd need to be a stereographic projection. Just a flat projection is fine. Maybe you had in mind constraining it to be a 1-to-1 map? In either case, for now let's skip it.

> What do you think about a conceptual question about projecting a 3d scene to a viewport on a flat device? The question would be something like: "Consider the function that maps coordinates in this 3d scene to the 2d device that you are viewing it on. Is this function linear (yes/no)?

That's a good question, but instead of just showing the picture you'd want to add some clarity on what the function is, or at the very least describe it as a map such that vanishing points exist, or something. It's perfectly possible to have a flat (linear) projection onto the viewport.

PR #411, opened 2024-03-21
Essence of Linear Algebra: Publish Change of Basis

kurtbruns:
This pull request:

- Publishes the change of basis lesson for review
- Note, Firefox display issue for SVGs is fixed for this lesson


public/content/lessons/2016/change-of-basis/index.mdx
```diff
@@ -7,4 +7,395 @@ video: P2LTAUO1TdA
...
+
+<Figure
+    image="./figures/alternate-system/AlternativeCoordinateSystem.light.svg" 
+/>
+
+So in effect, we are speaking different languages. We're all looking at the same vectors in space, but Jennifer uses different words and numbers to describe them.
+
+## The grid
+
+One quick word on how we're representing things here: When animating 2d space, we typically use a square grid square grid. But this grid is just a construct, a way to visualize our coordinate system, and so it depends on our choice of basis vectors. Space itself has no intrinsic grid.
```
3b1b, line 80:
```suggestion
One quick word on how we're representing things here: When illustrating 2d space, we typically use a square grid square grid. But this grid is just a construct, a way to visualize our coordinate system, so it depends on our choice of basis vectors. Space itself has no intrinsic grid.
```

public/content/lessons/2016/change-of-basis/index.mdx
```diff
@@ -7,4 +7,395 @@ video: P2LTAUO1TdA
...
+<Figure
+    image="./figures/translating-transformations/FromOursToHers.light.svg" 
+    width="600px"
+/>
+
+However, vectors aren't the only thing we describe using coordinates. For this next part, it's important that you're comfortable representing transformations with matrices, and that you know how matrix multiplication corresponds to composing successive transformations. Definitely pause and take a look at <LessonLink id="linear-transformations">chapter 3</LessonLink> and <LessonLink id="matrix-multiplication">chapter 4</LessonLink> if any of this feels uneasy.
+
+### Follow Rotation
+
+Consider some linear transformation, like a $90$ degree counterclockwise rotation. When we represent this with a matrix, we follow where the basis vectors $\hat{\imath}$ and $\hat{\jmath}$ each go. $\hat{\imath}$ ends up at the spot with coordinates $\left[\begin{array}{c} 0 \\ 1 \end{array}\right]$, and $\hat{\jmath}$ ends up at the spot with coordinates $\left[\begin{array}{c} -1 \\ 0 \end{array}\right]$, so those coordinates become the columns of our matrix.
```
3b1b, line 353:
```suggestion
Consider some linear transformation, like a $90$ degree counterclockwise rotation. In our coordinate system, we represent this with the matrix $\left[\begin{array}{c} 0 & -1 \\ 1 & 0 \end{array}\right]$. The first column tells us that $\hat{\imath}$ goes to $\left[\begin{array}{c} 0 \\ 1 \end{array}\right]$, and the second column tells us that $\hat{\jmath}$ goes to $\left[\begin{array}{c} -1 \\ 0 \end{array}\right]$.
```

3b1b (commented):
This looks good to me, the illustrations are well done.

Two quick notes: I always found it a potentially bad choice to have given a pi creature with a feminine name the color pink, since it sort of implies a gender for all the blue ones. I might suggest making Jennifer light blue.

Also, given the use of inverse matrices, it might be worth throwing in a quick note about how the inverse of M = [[a, b], [c, d]] is (1 / det(M))[[d, -b], [-c, a]].  That is, swap along one diagonal, negate the other, and divide by the determinant.

PR #416, opened 2024-04-17
Deleted duplicate paragraph

a-t-em:
This pull request:

- Removed a duplicate paragraph
- Left the rest of the code untouched


3b1b:
I think that paragraph is accomplishing a different goal from the previous one, emphasizing that it's not any one particular choice of a small change that matters, but instead what happens as that size approaches 0, and also what specifically is meant by "so small it doesn't matter."

PR #417, opened 2024-04-19
Essence of Linear Algebra: Publish Quick Eigen 

kurtbruns:
This pull request:

- Publishes the Quick Eigen article for review


public/content/lessons/2021/quick-eigen/index.mdx
```diff
@@ -7,4 +7,357 @@ video: e50Bj7jn9IQ
...
 credits:
 - Lesson by Grant Sanderson
+- Text adaptation by Kurt Bruns
 ---
+
+This is a video for anyone who already knows what eigenvalues and eigenvectors are, and who might enjoy a quick way to compute them in the case of 2x2 matrices. If you're unfamiliar with eigenvalues, take a look at the <LessonLink id="eigenvalues"> previous chapter</LessonLink> which introduces them.
+
+You can skip ahead if you just want to see the trick, but if possible we'd like you to rediscover it for yourself, so let's lay down a little background.
+
+As a quick reminder, if the effect of a linear transformation on a given vector is to scale it by some constant, we call that vector an eigenvector of the transformation, and we call the relevant scaling factor the corresponding eigenvalue, often denoted with the letter lambda.
```
3b1b, line 17:
```suggestion
As a quick reminder, if the effect of a linear transformation on a given vector is to scale it by some constant, we call that vector an eigenvector of the transformation, and we call the relevant scaling factor the corresponding eigenvalue, often denoted with the letter lambda, $\lambda$. 
```

public/content/lessons/2021/quick-eigen/index.mdx
```diff
@@ -7,4 +7,357 @@ video: e50Bj7jn9IQ
...
+
+That's a bit of a mouthful, but again, we're assuming this is review for anyone reading.
+
+The usual way to compute eigenvalues and how most students are taught to carry it out, is to subtract a variable lambda off the diagonals of a matrix and solve for when the determinant equals $0$. For example, when finding the eigenvalues of the matrix $\left[\begin{array}{ll} 3 & 1 \\ 4 & 1 \end{array}\right]$ this looks like:
+
+$$
+\operatorname{det}\left(\left[\begin{array}{cc}
+3-\lambda & 1 \\
+4 & 1-\lambda
+\end{array}\right]\right)=(3-\lambda)(1-\lambda)-(1)(4)
```
3b1b, line 45:
```suggestion
\end{array}\right]\right)=(3-\lambda)(1-\lambda)-(1)(4) = 0
```

public/content/lessons/2021/quick-eigen/index.mdx
```diff
@@ -7,4 +7,357 @@ video: e50Bj7jn9IQ
...
+
+3. (We'll get to this...)
+
+Before getting to the third fact, notice how you can essentially read these first two values out of the matrix. Take this matrix $\left[\begin{array}{ll}8 & 4 \\ 2 & 6\end{array}\right]$ as an example. Straight away you can know that the mean of the eigenvalues is the same as the mean of $8$ and $6$, which is $7$.
+
+$$
+m = \frac{\lambda_1 + \lambda_2}{2} = 7
+$$
+
+Likewise, most linear algebra students are well-practiced at finding the determinant, which in this case is $48 - 8$, so you know the product of our two eigenvalues is $40$.
```
3b1b, line 94:
```suggestion
Likewise, most linear algebra students are well-practiced at finding the determinant, which in this case is $8 \cdot 6 - 4 \cdot 2 = 48 - 8$, so you know the product of our two eigenvalues is $40$.
```

3b1b:
One bigger change I might recommend is to deviate from the video structure and put the trick and exercise up at the top of the article.

- Show the trick
    - Given a matrix [[a, b], [c, d]], even if you can't immediately see its eigenvalues, you can almost read off the mean and product of those eigenvalues:
        - Mean = mean(diagonal) = (a + d) / 2
        - Product = det = ad - bc
    - Given the mean and product of two numbers, you can recover them with m±sqrt(m^2 - p).  [Acknowledge that this is essentially the quadratic formula]
- Give the multiple-choice questions
- Explain why the trick works
    - Review the usual way to compute them by subtracting lambda off diagonals, expanding the det, and solving the characteristic polynomial.
    - Facts about det, trace, etc.
- QM example

3b1b:
Also, while reading this, I noted a slightly faster way. Let's write a general matrix as [[a, b], [c, d]], with the specific example [[2, 4], [6, 8]] to illustrate

1. Write a, d as m ± s. For example, 2, 8 = 5 ± 3
2. The eigenvalues are m ± sqrt(s^2 + bc). For example 5 ± sqrt(9 + 4 * 6) = 5 ± sqrt(33)

I'm hesitant to add much more because the trick doesn't really matter, its the exercise in thinking about why it works that matters.

public/content/lessons/2021/quick-eigen/index.mdx
```diff
@@ -7,4 +7,357 @@ video: e50Bj7jn9IQ
...
+& = 7 \pm \sqrt{9} \\ \rule{0pt}{1.25em}
+& = 4, 10
+\end{aligned}
+$$
+
+</Question>
+
+## Last thoughts
+
+This is an oddly specific trick, but it's something that could be helpful in college, so if you know any students you think might benefit from this, consider sharing it with them.
```
3b1b, line 347:
```suggestion
```

I'd happily remove this line.

PR #422, opened 2024-05-07
Essence of Linear Algebra: Publish Abstract Vector Spaces

kurtbruns:
This pull request:

- Publishes the Abstract Vector Spaces article for review


public/content/lessons/2016/abstract-vector-spaces/index.mdx
```diff
@@ -7,4 +7,248 @@ video: TgKwz5Ikpc8
...
+A transformation is *linear* if it satisfies two properties: Additivity and scaling.
+
+Additivity means that when you add any two vectors $\vec{\mathbf{v}}$ and $\vec{\mathbf{w}}$, then apply the transformation to their sum, you get the same result as if you add the transformed versions of $\vec{\mathbf{v}}$ and $\vec{\mathbf{w}}$.
+
+<Figure
+    image="./figures/standard-definition-of-linear/AdditivityProperty.light.svg" 
+    width="520px"
+/>
+
+The scaling property is that when you scale a vector $\vec{\mathbf{v}}$ by some number, then apply the transformation to the result, you get the same ultimate vector as if you scale the transformed version of $\vec{\mathbf{v}}$ by that same amount.
```
3b1b, line 106:
```suggestion
The second property says that when you scale a vector $\vec{\mathbf{v}}$ by some number $s$, then apply the transformation to the result, you get the same ultimate vector as if you scale the transformed version of $\vec{\mathbf{v}}$ by that same amount $s$.
```

public/content/lessons/2016/abstract-vector-spaces/index.mdx
```diff
@@ -7,4 +7,248 @@ video: TgKwz5Ikpc8
...
+/>
+
+Take a moment to imagine yourself as a mathematician developing the theory of linear algebra, you want all your definitions and discoveries to work for all of these vector-ish things in full generality.
+
+<Figure
+    image="./figures/abstract-vector-spaces/YouAsAMathematician.light.svg"
+    width="520px"
+/>
+
+These sets of vector-ish things, like arrows, lists of numbers and functions, are called “vector spaces”, and what you as the mathematician might do is say “Hey everyone, I don't want to think about all the different types of crazy vector spaces that y'all might come up with”, so you establish a list of rules that vector addition and scaling have to abide by in order for the results that you-the-mathematician discover to apply to someone's chosen definition of vectors.
```
3b1b, line 212:
```suggestion
A set of vector-ish objects, like the set of all points in 3d space, the set of all lists of 4 numbers, the set of all finite polynomial functions, etc., is called a “vector space”. What you as the mathematician might do is say “Hey everyone, I don't want to think about all the different types of crazy vector spaces that y'all might come up with”, so you establish a list of rules that vector addition and scaling have to abide by in order for the results that you-the-mathematician discover to apply to someone's chosen definition of vectors.
```

public/content/lessons/2016/abstract-vector-spaces/index.mdx
```diff
@@ -7,4 +7,248 @@ video: TgKwz5Ikpc8
...
+
+<Figure
+    image="./figures/abstract-vector-spaces/PiCreatureSpace.light.svg" 
+/>
+
+You as the mathematician never have to think about all the possible crazy vector spaces people might define; you just have to prove your results in terms of these axioms, so anyone whose definitions satisfy these axioms can happily apply your results. As a consequence, you'd tend to phrase all your results abstractly, which is to say only in terms of these axioms, rather than centering on a specific type of vector like arrows in space or functions.
+
+For example, this is why every textbook you find will define linear transformations in terms of additivity and scaling, rather than talking about grid lines remaining parallel and evenly spaced, even though the latter is more intuitive and, at least in my view, more helpful for first-time learners.
+
+So the mathematician's answer to “what are vectors?” is to ignore the question. In the modern theory, the form they take doesn't really matter: arrows, lists of numbers, functions, pi creatures, really they can be *anything*, so long as there is some notion of adding and scaling vectors that follows these rules.
```
3b1b, line 238:
```suggestion
So the mathematician's answer to “what are vectors?” is to ignore the question. In modern theory, the form they take doesn't matter: arrows, lists of numbers, functions, pi creatures, really they can be *anything*, so long as there is some notion of adding and scaling vectors that follows these rules.
```

public/content/lessons/2016/abstract-vector-spaces/index.mdx
```diff
@@ -7,4 +7,248 @@ video: TgKwz5Ikpc8
...
+
+It's like asking what the number 3 really is. Whenever it comes up concretely, it's in the context of some triplet of things. But in math it's treated as an abstraction of all possible triplets of things that lets you reason about all those triplets using a single idea.
+
+<Figure
+    image="./figures/abstract-vector-spaces/Triplets.light.svg" 
+/>
+
+Some goes with vectors, which have many embodiments, but math abstracts them all into a single intangible notion of a vector space. As anyone reading this series knows, we think it's better to begin reasoning about vectors in a concrete, visualizable setting like 2d space with arrows rooted at the origin. But as you learn more linear algebra, know that these tools apply much more generally, and that this is the underlying reason why textbooks and lectures tend to be phrased, well, abstractly.
```
3b1b, line 246:
```suggestion
It's the same with vectors, which have many embodiments, but math abstracts them all into a single intangible notion of a vector space. As anyone reading this series knows, we think it's better to begin reasoning about vectors in a concrete, visualizable setting like 2d space with arrows rooted at the origin. But as you learn more linear algebra, know that these tools apply much more generally and that this is the underlying reason why textbooks and lectures tend to be phrased, well, abstractly.
```

3b1b (commented):
Looking good, just a few tweaks in this review. Thanks!

PR #424, opened 2024-05-20
Straightforward example to puzzle

JJones780:
A more applicable example in the final puzzle introducing non-square matrices. The new example is also better related to the puzzle question. Also caught a few typos -  hopefully didn't introduce any.
I'm not confident in my use of subscripts in the matrices as they might be too different from the existing format?

### DIFF
--- a/public/content/lessons/2016/3d-transformations/index.mdx
+++ b/public/content/lessons/2016/3d-transformations/index.mdx
@@ -378,25 +378,30 @@ Here's another puzzle for you: It's also meaningful to talk about a linear trans
 
 A simple example from 3d to 2d would be the shadow cast by a 3d object onto a 2d plane. Although, in order for the transformation to be linear, the light rays should be considered parallel to eachother to simplify the problem. If you consider the light source to be something really far away, like the sun, then this is a reasonable choice to make.
 
-You can represent these transformations with matrices. The number of columns corresponds to the dimension of the input and the number of rows corresponds to the dimension of the output. For example, a matrix that maps coordinates on a sphere to a plane would have three columns and two rows.
+You can represent these transformations with matrices. The number of columns corresponds to the dimension of the input and the number of rows corresponds to the dimension of the output. For example, a matrix that maps coordinates in 3d to 2d would have three columns and two rows.
 
 $$
-A = \left[\begin{array}{ccc}
-a_1 & a_2 & a_3 \\
-a_4 & a_5 & a_6
+M = \left[\begin{array}{ccc}
+m_{ii} & m_{ij} & m_{ik} \\
+m_{ji} & m_{jj} & m_{jk}
 \end{array}\right]
 $$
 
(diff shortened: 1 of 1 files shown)


3b1b:
Thanks for the contribution. The typo correction is appreciated, but the indices don't quite make sense. Typically you'll use variable subscripts like i, j, k not for specific entries, but to write a generic expression for all entries.

JJones780:
I've changed the indices to 1,2,3..  
If the example is not wanted I can pare down the pull request to the typos only, or you can make those changes independently and close this.

I really appreciate your videos  - I knew that I had missed out on some intuitive understanding back when I studied these topics and it's a great feeling to revisit even the simpler math and really connect to it intuitively.

public/content/lessons/2016/3d-transformations/index.mdx
```diff
@@ -378,25 +378,30 @@ Here's another puzzle for you: It's also meaningful to talk about a linear trans
...
-a_1 & a_2 & a_3 \\
-a_4 & a_5 & a_6
+M = \left[\begin{array}{ccc}
+m_{1} & m_{2} & m_{3} \\
+m_{4} & m_{5} & m_{6}
 \end{array}\right]
 $$
 
-It's meaningful to talk about multiplying these matrices when the number of columns on the left matrix is equal to the number of rows on the right matrix. That whey when apply these matrices to a vector, reading right to left, the dimensions of the input and output match up.
+It's meaningful to talk about multiplying these matrices when the number of columns on the left matrix is equal to the number of rows on the right matrix. That when we apply these matrices to a vector, reading right to left, the dimensions of the input and output are as we expect.
```
3b1b, line 390:
```suggestion
It's meaningful to talk about multiplying these matrices when the number of columns on the left matrix is equal to the number of rows on the right matrix. That way, when we apply these matrices to a vector, reading right to left, the dimensions of the input and output are as we expect.
```

public/content/lessons/2016/3d-transformations/index.mdx
```diff
@@ -378,25 +378,30 @@ Here's another puzzle for you: It's also meaningful to talk about a linear trans
...
+m_{4} & m_{5} & m_{6}
+\end{array}\right]
+\left[\begin{array}{c}
+a_1  \\
+a_2  \\
+a_3
+\end{array}\right]
+=\left[\begin{array}{c}
+b_1 \\
+b_2
```
3b1b, line 404: You should leave this as it originally was, from the context it looks like the intent was to display two matrices being multiplied by each other, not to show a matrix-vector product example.

JJones780:
I put the original matrices back in place.  Just the typos fix including your last suggestion.
Should be good to go.

public/content/lessons/2016/3d-transformations/index.mdx
JJones780: done

public/content/lessons/2016/3d-transformations/index.mdx
JJones780: all fixed

Issue #407, opened 2024-02-28
Firefox SVG Rendering


kurtbruns:
Firefox renders some of the SVGs in the Linear Algebra Series differently from Safari and Chrome, sometimes resulting in a rather poor output image. See as an example the lesson: https://www.3blue1brown.com/lessons/linear-transformations

Example 1:

<img width="506" alt="Screenshot 2024-02-28 at 3 16 31 PM" src="https://github.com/3b1b/3Blue1Brown.com/assets/12160501/a99e8f4e-d582-40cc-9e71-f6f7bcc06a18">

Example 2:

<img width="560" alt="Screenshot 2024-02-28 at 3 16 46 PM" src="https://github.com/3b1b/3Blue1Brown.com/assets/12160501/055484fc-af18-4d35-b88c-ce64ae781f65">

PR #426, opened 2024-05-23
Essence of linear algebra edits

kurtbruns:
This pull request:

- Fixes images not rendering properly in Firefox in the Essence of Linear Algebra Series #407


kurtbruns (commented):
Double checked each image using the image diff in the code review.

PR #453, opened 2024-12-15
Transformers Blog Post

JustinGamer191:
Pull Request for Transformers Blog Post.


public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+### Embedding Matrix
+
+The model has a predefined vocabulary, some list of all possible words, say 50,000 of them, and the first matrix of the transformer, known as the **embedding matrix**, will have one column for each of these words. These columns determine what vector each word turns into in that first step. We label it as $W_E$, and like all the matrices we see, its values begin random, but they're going to be learned based on data.
+
+Turning words into vectors was common practice in machine learning long before transformers, and it sets the foundation for everything that follows, so let’s take a moment to get familiar with it. This is often called _embedding the word_, which invites thinking of these vectors geometrically as points or directions in some space. Visualizing a list of three numbers as coordinates for a point in 3D space is no problem, but word embeddings tend to be very high dimensional. For GPT-3, they have 12,288 dimensions, and as seen, it matters to work in a space with lots of distinct directions.
```
3b1b, line 103:
```suggestion
Turning words into vectors was common practice in machine learning long before transformers. This is often called _embedding the word_, which invites thinking of these vectors geometrically as points or directions in some space. Visualizing a list of three numbers as coordinates for a point in 3D space is no problem, but word embeddings tend to be very high dimensional. For GPT-3, they have 12,288 dimensions, and as you will see, it matters to work in a space with lots of distinct directions.
```

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+<Figure image="embedding.png" width="100%"/>
+
+In the same way that you can take a 2D slice through 3D space and project points onto that slice, for the sake of visualizing word embeddings, we will do something analogous by choosing a 3D _slice_ through the very high-dimensional space, projecting word vectors onto that, and displaying the result.
+
+### Direction
+
+The big idea we need to understand here is that as a model tweaks and tunes its weights to decide how exactly words get embedded as vectors during training, it tends to settle on a set of embeddings where directions in this space have semantic meaning. Below, a simple word-to-vector model is running, and when I run a search for all words whose embeddings are closest to that of _tower_, they all generally have the same vibe.
```
3b1b, line 111:
```suggestion
The big idea we need to understand here is that as a model tweaks and tunes its weights to decide how exactly words get embedded as vectors during training, it tends to settle on a set of embeddings where directions in this space have meaning. Below, a simple word-to-vector model is running, and when I run a search for all words whose embeddings are closest to that of _tower_, they all generally have the same vibe.
```

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+<Figure image="tower.png" width="100%"/>
+
+Another classic example of this is when the difference between the vectors for _woman_ and _man_ is taken, which can be visualized as a vector in this space connecting the tip of one to the tip of the other. This difference is quite similar to the difference between _king_ and _queen_. So, if the word for a female monarch was unknown, it could be found by taking _king_, adding the direction of _woman_ minus _man_, and searching for the closest word embedding.
+
+<Figure image="queen.png" width="100%"/>
+
+At least, kind of. Despite this being the classic example, for this model the true embedding of _queen_ is a little farther off than the difference would suggest, presumably because the way _queen_ is used in training data is not merely a feminine version of a king. Family relations illustrate the idea better:
```
3b1b, line 119:
```suggestion
At least, kind of. Despite this being the classic example, for the model I was playing with when making this video, the true embedding of _queen_ is a little farther off than the difference would suggest, presumably because the way _queen_ is used in training data is not merely a feminine version of a king. Family relations illustrate the idea better:
```

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+<Figure image="gender.png" width="100%"/>
```
3b1b, line 121: This image is way too busy. Just include the thee side-by-side examples, no need for added arrows or annotations.

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+The idea here is that it seems as if, during training, the model found it advantageous to choose embeddings such that one direction in this space encodes gender information.
+
+Another example of this would be taking the embedding of _Italy_, subtracting the embedding of _Germany_, and adding it to the embedding of _Hitler_. This results in something very close to the embedding of _Mussolini_. It’s as if the model learned to associate some directions with Italian-ness and others with WWII Axis leaders.
+
+<Figure image="italy.png" width="100%"/>
```
3b1b, line 127: Remove the added red lines and arrows. The function of those red lines is to highlight the word "direction" when it's being spoken in narration, but in a still image, they only add distraction.

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+### Dot Product
+
+One bit of mathematical intuition helpful to have in mind as we continue is how the dot product of two vectors can be thought of as measuring how well they align. Computationally, dot products involve multiplying all aligning components and adding the result. Geometrically, the dot product is positive when the vectors point in a similar direction, zero if they're _perpendicular_, and negative when they point in opposite directions.
```
3b1b, line 131: Add a quick image or LaTeX expression to show the computation of a dot product.

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+<Figure image="italy.png" width="100%"/>
+
+### Dot Product
+
+One bit of mathematical intuition helpful to have in mind as we continue is how the dot product of two vectors can be thought of as measuring how well they align. Computationally, dot products involve multiplying all aligning components and adding the result. Geometrically, the dot product is positive when the vectors point in a similar direction, zero if they're _perpendicular_, and negative when they point in opposite directions. 
+
+For example, suppose we wanted to test if the embedding of _cats_ minus that of _cat_ represents a kind of plurality direction in this space. To test this, notice what values we get when we take a dot product between this vector and various singular and plural nouns. It looks like the plural ones do indeed end up with consistently higher values than singular ones. Also, if we take this dot product with the embeddings of the words _one_, _two_, _three_, and so on, they give increasing values, as if it’s quantitatively measuring how plural the model finds a given word.
```
3b1b, line 133:
```suggestion
For example, suppose we wanted to test if the embedding of _cats_ minus that of _cat_ represents a kind of plurality direction in this space. To test this, you could take the dot product between this vector and various singular and plural nouns. When I did this for a simple word-to-vector model while making this video, it looked like the plural ones do indeed end up with consistently higher values than singular ones. Also, taking this dot product with the embeddings of the words _one_, _two_, _three_, and _four_, they give increasing values, as if it’s quantitatively measuring how plural the model finds a given word.
```

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+<Figure image="dotProduct.png" width="100%"/>
+
+Again, how specifically each word gets embedded is learned using data. The embedding matrix, whose columns store the embedding of each word, is the first pile of weights in our model. Using the GPT-3 numbers, the vocabulary size is 50,257, and again, technically this consists not of words, per se, but different little chunks of text called tokens. The embedding dimension is 12,288, giving us **617,558,016 weights** in total for this first step. Let’s go ahead and add that to a running tally, remembering that by the end we should count up to 175 billion weights.
```
3b1b, line 137:
```suggestion
Again, how specifically each word gets embedded is learned using data. The embedding matrix, whose columns store the embedding of each word, is the first pile of weights in our model. Using the GPT-3 numbers, the vocabulary size is 50,257, and again, technically this consists not of words, per se, but different little chunks of text called tokens. The embedding dimension is 12,288, giving us **617,558,016 weights** in total for this first step. Let’s go add that to a running tally, remembering that by the end we should count up to 175 billion weights.
```

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+<Accordion title={
```
3b1b, line 139: Again, use the <Question> widget instead

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+    <>
+        Comprehension Question 3: In a vector space representing meanings, the following relationship holds: <br />
```
3b1b, line 141:
```suggestion
        Comprehension Question 3: In a vector space representing meanings, which of the following would you expect to be closest to the vector Embedding("Cat") - Embedding("Meow") + Embedding("Bark")<br />
```

And format the answers to also be Embedding("Dog"), etc.

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+    </>
+} 
+children={
+    "In this vector space, Cat - Meow represents a sort of sound direction. The animal that most commonly makes the bark sound is C. Dog"
+    }
+/>
+
+### Beyond words
+
+In the case of a transformer, we also want to think of the vectors in this embedding space as not merely representing individual words. These embeddings will also encode information about the position of the word, which we’ll talk about later. But more importantly, we should view these vectors as having the capacity to soak in context.
```
3b1b, line 157:
```suggestion
In the case of a transformer, we also want to think of the vectors in this embedding space as not merely representing individual words. For one thing, these embeddings will also encode information about the position of the word, but more importantly, they need to have the capacity to soak in context.
```

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+<Figure image="king.png" width="100%"/>
+
+Think about our understanding of a word, like quill. Its meaning is clearly informed by its surroundings and context, whether it be a hedgehog quill or a type of pen. Sometimes, we may even include context from a long distance away. When putting together a model that is able to predict the next word, the goal is to somehow empower it to do the same thing: take in context efficiently. 
+
+In that very first step, when you create the array of vectors based on the input text, each one is simply plucked out of this embedding matrix, and each one only encodes the meaning of the single word it's associated with; no input from its surroundings. But as these vectors flow through the network, we should view the primary goal of this network as enabling each one of those vectors to soak up meaning that is more rich and specific than what mere individual words can represent.
```
3b1b, line 165:
```suggestion
In that very first step, when you create the array of vectors based on the input text, each one is simply plucked out of this embedding matrix, and each one only encodes the meaning of the single word it's associated with. It's effectively a lookup table with no input from the surroundings. But as these vectors flow through the network, we should view the primary goal of this network as enabling each one of those vectors to soak up meaning that is more rich and specific than what mere individual words can represent.
```

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+The network can only look at a fixed number of vectors at a time, known as its _context size_, which for GPT-3 is 2048. So the data flowing through the network will look like this array of 2048 columns, each of which has around 12k dimensions. This context size limits how much text the transformer can incorporate to make its prediction of the next word, which is why long conversations with the early versions of ChatGPT often gave the feeling of the bot losing the thread of conversation.
```
3b1b, line 167:
```suggestion
The network can only look at a fixed number of vectors at a time, known as its _context size_. GPT-3 was trained with a context size of 2048 tokens. So the data flowing through the network will look like this array of 2048 columns, each of which has around 12k dimensions. This context size limits how much text the transformer can incorporate to make its prediction of the next word, which is why long conversations with the early versions of ChatGPT often gave the feeling of the bot losing the thread of conversation.
```

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+<Figure image="InitialContext.png" width="100%"/>
+
+## Unembedding
+
+We'll go into the details of the Attention Block in due time, but first we'll skip ahead and talk about what happens at the very end of the Transformer. Remember, the desired output is a probability distribution over all possible chunks of text that might come next.
```
3b1b, line 173:
```suggestion
We'll go into the details of the Attention Block in the next chapter, but first we'll skip ahead and talk about what happens at the very end of the Transformer. Remember, the desired output is a probability distribution over all possible chunks of text that might come next.
```

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+This is often called the **unembedding matrix**, and we'll give it the label $$W_U$$.  Again, like all the weight matrices we’ll see, its entries begin random but are tuned during the training process.
+
+<Figure image="WU.png" width="100%"/>
+
+Keeping score on our total number of parameters, the unembedding matrix has one row for each word in the vocabulary, giving 50,257 words, and each row has the same number of elements as the dimension of the embedding, giving 12,288 columns. It’s very similar to the embedding matrix, just with the dimensions of the rows and columns swapped, so it adds another 617M parameters to the network, making our parameter count so far a little over a billion; a small but not insignificant fraction of the 175 billion that we'll end up with in total.
+
+<Accordion title={
```
3b1b, line 192: Again, use the <Question> Widget

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+children={
+    "C. 32,000 rows. Since the dimension the embedding matrix exists in is 10,000, the embedding matrix must have 10,000 rows. Then the embedding matrix must have 32,000 columns, one for each token. The unembedding matrix is similar to the embedding matrix, but the dimensions of the rows and columns are swapped, giving 32,000 rows and 10,000 columns."
+    }
+/>
+
+## Softmax
+
+For the last lesson of this chapter, we will go over the softmax function, since it will appear again once we dive into the attention blocks.
+
+The idea is that if we want a sequence of numbers to serve as a probability distribution, say a distribution over all possible next words, all the values should be between 0 and 1 and should all add up to be 1. However, in deep learning, where everything we do looks like a matrix-vector product, the outputs we get by default won’t abide by this at all. The values are often negative or sometimes greater than 1, and they almost certainly don’t all add up to 1.
```
3b1b, line 210:
```suggestion
The idea is that if we want a sequence of numbers to serve as a probability distribution, say a distribution over all possible next words, all the values should be between 0 and 1 and should all add up to be 1. However, in deep learning, where so much of what we do looks like a matrix-vector products, the outputs we get by default won’t abide by this at all. The values are often negative or sometimes greater than 1, and they almost certainly don’t all add up to 1.
```

public/content/lessons/2024/gpt/index.mdx
```diff
@@ -1,9 +1,231 @@
...
+
+The reason for calling it _softmax_ is that instead of simply pulling out the biggest value, it produces a distribution that gives weight to all the relatively large values, commensurate with how large they are. If one entry in the input is much bigger than the rest, the corresponding output will be very close to 1, so sampling from the distribution is likely the same as just choosing the maximizing index from the input.
+
+<Figure image="howsoftmax.png" width="100%"/>
+
+In some cases, such as when ChatGPT is using this distribution to sample each next word, we can add a little extra spice into this softmax function with a constant _T_ inserted into all these exponents. It's called _T_ since it vaguely resembles the role of temperature in certain thermodynamics equations. The effect of this constant is that when _T_ is larger, more weight is given to lower values, making the distribution more uniform. When _T_ is smaller, the larger values will dominate the distribution, and in the extreme setting when _T_ is equal to 0, all the weight goes to the maximum value.
```
3b1b, line 222:
```suggestion
In some cases, such as when ChatGPT is using this distribution to sample each next word, we can add a little extra spice into this softmax function with a constant _T_ inserted into all these exponents. We call it "temperature" since it vaguely resembles the role of temperature in certain thermodynamics equations. The effect of this constant is that when _T_ is larger, more weight is given to lower values, making the distribution more uniform. When _T_ is smaller, the larger values will dominate the distribution, and in the extreme setting when _T_ is equal to 0, all the weight goes to the maximum value.
```

public/content/lessons/2024/gpt/index.mdx
JustinGamer191: Sorry, will change all the mcqs.

PR #564, opened 2026-03-28
Refresh

### LINKED ISSUE #280: Scrolling up on the Newton's fractal interactive app zooms in all the way
**Describe the bug**
On the interactive [here](https://www.3blue1brown.com/lessons/newtons-fractal), scrolling up zooms the view all the way in, rather than zooming in slowly.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

### LINKED ISSUE #408: Free Response Edit
Feature:

From Grant: "As a side note, I think we should edit the free response question to not have a textbox since it's hard to envision anyone literally typing for anything more than an answer with a few characters. It should instead be labeled something like "Pause and ponder", with the same blurred-out result with a button to reveal."

### LINKED ISSUE #456: The 3d-surface-interactive part of inscribed-rect-v2 missing ParametricGeometry
**Describe the bug**
The interactive 3d surface view on https://www.3blue1brown.com/lessons/inscribed-rect-v2 is not working. It looks like you might be missing the extension for ParametricGeometry in THREE.

**To Reproduce**
Go to the page, scroll to the interactive viewer and it doesn't work.

BTW - source code for the video doesn't appear ot have been pushed, figure you might just be enjoying the holidays :) https://github.com/3b1b/videos/tree/master/_2024/inscribed_rect

vincerubinetti:
This PR is a redesign and rewrite of the entire website, spawned from an [improvement plan](https://github.com/user-attachments/files/26320009/3b1b.site.refresh.docx) I made years ago.

**Overall approach**

- Ruthlessly simplify and prune features. The best feature is no feature.
- No git LFS. Simple file management.
- No barrel files or folders. Descriptive filenames over "index.js".
- Type safety at editor/build level, not at runtime. No PropTypes.
- Increased robustness and consistency with tests, linting, and auto-formatting.
- Prioritize accessibility.
- Split by domain not type. Example: /blog/images and /blog/components rather than /images/blog and /components/blog.
- Explicit, type-safe imports (idiomatic). Avoid "blind" string references to /public/content.
- Consolidate content, simplify writing.

**Important changes**

Some of these may be controversial, see improvement plan for more discussion, and continue discussion in this PR.

- Next → React Router
- Lesson assets no longer in repo, expected to be in cloud bucket
- TypeScript
- Yarn → Bun
- Pre-rendered KaTeX → client-rendered MathJAX
- Linting w/ ESLint
- Formatting w/ Prettier
- Styling w/ Tailwind, no more Sass
- End-to-end tests w/ Playwright
- Accessibility tests w/ Deque Axe
- React 19
- YAML → type-safe JSON
- FontAwesome icons → Phosphor Icons
- Reorganize repo structure
- Convert all lessons to new components
- Lodash and React hook libraries to simplify code

**Medium changes**

- Migrate assets to GCP and compress
- Aesthetic refresh
- Rework copy text on "main" pages (home, about, etc.) but leave more "personal" pages (blog posts, lessons, talent pages) alone
- Use modern CSS techniques
- Add dark mode toggle
- Fuzzy-search lessons
- Lesson search in header
- Remove dead code
- Clean up components
- Simplify old code
- Fix broken interactives
- Rework and simplify shader interactives
- Feedback form
- Add JSON LD metadata on lessons and blog posts for better SEO
- Bunch of useful util funcs
- Add fun flare like hilbert viz and pi confetti
- Tests/checks for broken links, unused imports, asset size, spelling, and more
- Many bug fixes by virtue of redoing architecture
- Fix typos in entire site including lessons and asset file names

**More tasks**

https://docs.google.com/document/d/1bN8tClQyDXNkcqdbh8hbxKdT8a6yJZf4p9PKyJ9vyrk


vincerubinetti:
Notes to reviewers:

- You may have to run the site locally to preview the changes. Netlify seems to be configured for Next.js through the web UI, which is making this PR preview fail. [And apparently you can't disable/override that through a branch `netlify.toml` config unless you make a `netlify.toml` on `main` to specify things instead of through the web UI](https://claude.ai/share/38603473-ec63-4821-8287-6c0eb0a96d5e). Either way, once merged, future PR previews will work though.
- ✅ ~When ready to merge, I'll do a `git merge origin/main -s ours` to resolve the merge conflicts and make the new `main` exactly reflect what's in this branch. I have attempted to capture all of the PRs that have been made since I branched this off of `main`, e.g. adding Epoch AI.~

app/components/MathJax.tsx
vincerubinetti: I just noticed that display math wasn't rendering, I'm assuming what your MathJax edits were trying to address.

But I believe the true issue is with this line. I added it in 56395154e4811db0ee653b7a1922d373458aa89a, probably intending `contents` ("hide" the `<pre>` wrapper but still show its children) instead of `sr-only` (effectively visually hiding `<pre>` completely).

app/components/MathJax.tsx
```diff
@@ -112,7 +112,9 @@ const render = async () => {
       content = await window.MathJax.tex2svgPromise?.(math, { display });
     }
     if (!content) continue;
-    // insert content
-    element.replaceWith(content);
+    // insert content — for display math, insert after the <pre> (which is now sr-only);
+    // for inline math, replace the <code> in place
+    if (display) parent.insertAdjacentElement("afterend", content);
+    else element.replaceWith(content);
```
vincerubinetti: Have to be very careful with this. It took a lot to get this MathJax stuff right, and it'll affect the entire site including lessons. I wish it wasn't so fragile, but I still think the benefits of client loading vs pre-rendering and mathjax vs katex are still worth it.

I believe this has the potential to mess with React hydration. This is adding a new element which could throw off React's ability to track the order of elements in lists. I ran into this before trying to do a similar thing, and it made the table of contents crash the app sometimes. Replacing the content (a node that react shouldn't have a reason to touch) and not adding/removing dom nodes seems to be stable.

vite.config.ts
```diff
@@ -73,6 +73,23 @@ export default defineConfig(() => ({
...
+            children: [{ type: "element", tagName: "code", properties: { className: ["language-math"] }, children: [{ type: "text", value: node.value }] }],
+          }),
+          inlineMath: (_state, node) => ({
+            type: "element",
+            tagName: "code",
+            properties: { className: ["language-math"] },
+            children: [{ type: "text", value: node.value }],
+          }),
+        },
+      },
```
vincerubinetti: I don't think any of this should be necessary, it seems to maybe just have overriden the mistake i made by making `<pre>` `sr-only`.  I believe the `remarkMath` plugin already does something similar to this by default.

app/pages/talent/Partners.tsx
```diff
@@ -16,50 +14,41 @@ const order = [
...
             >
-              <div className="flex size-full items-center gap-12 p-4 transition group-hocus:opacity-0 max-md:flex-col">
-                <img src={logo} alt="" className="w-40" />
-                <div className="flex flex-col gap-4 text-left font-sans max-md:items-center max-md:text-center">
-                  <div className="text-xl font-bold">{name}</div>
-                  <div className="text-lg">{tagline}</div>
-                </div>
+              <img src={logo} alt="" className="h-32 w-full object-contain" />
+              <div className="mt-2 font-sans text-2xl font-bold">{name}</div>
+              <div className="text-center text-sm text-gray opacity-0 transition group-hocus:opacity-100">
```
vincerubinetti: I'd recommend almost never using `text-sm`, for accessibility reasons.

app/util/math.ts
```diff
@@ -6,6 +6,26 @@ export const sin = (degrees: number) => Math.sin(tau * (degrees / 360));
 export const cos = (degrees: number) => Math.cos(tau * (degrees / 360));
 export const tan = (degrees: number) => Math.tan(tau * (degrees / 360));
 
+// simple seeded PRNG (mulberry32)
+export function seededRandom(seed: number) {
```
vincerubinetti: There are some convenient JS libraries for this, and lodash has a shuffle method, but unfortunately it's a burden to get them to work together, so probably cleaner to leave this as is.

Also completely inconsequential (in 99% of cases) but I've used a pattern of `function () {}` syntax for React components and `() => {}` (arrow) syntax for everything else.

app/pages/talent/Partners.tsx
```diff
@@ -46,14 +46,16 @@ export default function Partners() {
...
+          const logoDark = getLogoDark(id);
           return (
             <Link
               key={id}
               to={href("/talent/:id", { id })}
               arrow={false}
               className="group flex flex-col items-center gap-2 rounded-md p-4 text-black no-underline hocus:bg-theme/15"
             >
-              <img src={logo} alt="" className="h-32 w-full object-contain" />
+              <img src={logo} alt="" className={`h-32 w-full object-contain${logoDark ? " dark:hidden" : ""}`} />
```
vincerubinetti: Use `clsx`, a util func for combining class names, as a more ergonomic and flexible alternative to string concat. Unfortunately copilot has been really bad about auto-completing with `clsx` even though it is used in every `className` combo throughout the code base.

For the switching, the darkmode state is a reactive var we can just use like `useAtomValue(darkModeAtom)`.

app/pages/talent/Partners.tsx
```diff
@@ -16,50 +14,41 @@ const order = [
...
-          const { name = "", tagline = "", quote = "" } = partner.frontmatter;
+          const { name = "", tagline = "" } = partner.frontmatter;
           const logo = getLogo(id) ?? "";
           return (
-            <Card
+            <Link
               key={id}
               to={href("/talent/:id", { id })}
-              className="group relative bg-[white]! text-[black]!"
+              arrow={false}
```
vincerubinetti: Not needed because arrow not added by default when link is internal. If one of these links went directly to a partner's site then it'd be needed, but that doesn't seem to be an intention.

app/pages/talent/Partners.tsx
vincerubinetti: I'm also thinking we should just show the tagline all the time? I think hover effects should mostly be used for fun/flare, and less for critical information. These feel like fairly important for discovery? Also remember that touch devices don't have hover in the same way mouse devices do.

app/components/Feedback.tsx
```diff
@@ -118,83 +118,90 @@ export default function Feedback() {
...
+                          <dt>{key}</dt>
+                          <dd>{value}</dd>
+                        </Fragment>
+                      ))}
+                    </dl>
+                  </Tooltip>
+                  . You'll get a link to it once it's created, where you can
+                  attach screenshots or more details. For general inquiries, see
+                  the{" "}
+                  <Link to={`${href("/about")}#faqs`} onClick={close}>
```
vincerubinetti: I'll just have the dialog component close on any url pathname change. I can't think of a scenario/precedence where this wouldn't be the desired behavior anyway.

PR #587, opened 2026-04-21
Showcase translations

3b1b:
This is a test, just to brainstorm one way to better showcase translated content. I'm not sure if it's the way to go, because it adds complexity, but the idea is to integrate it with the home page and make it possible to filter for videos based on what languages they are available in.


vincerubinetti:
You may have seen already, but apparently the YouTube widget currently doesn't expose any way to set the language of the audio track. You can only set the language of the widget interface (and I guess captions?).

And FYI in the `<youtube-video>` video component, [here are the attributes we can set on it](https://github.com/muxinc/media-elements/blob/main/packages/youtube-video-element/youtube-video-element.d.ts). The package defines these types but they don't get picked up by TypeScript and associated with the `<youtube-video>` element; I'll try to set that up somehow. But for now, you can do e.g. `<youtube-video config={{ hl: "en" }}>` to set the interface language or [any of the other params](https://developers.google.com/youtube/player_parameters#Parameters).

3b1b:
Yeah, it's a shame. In _theory_ I think YouTube serves an audio track based on a users previous established preferences (not just what they set, but also from past behavior). Having an explicit "how to switch audio tracks" message doesn't feel great, but it's not terrible in this case.

PR #594, opened 2026-04-25
Update Agents.md

### LINKED ISSUE #591: Add to claude.md, replace readme.md
#590 adds a `claude.md` file. There's a lot of overlap with the readme currently. We can probably just replace the readme with something very minimal or different, and point to `claude.md` for all the technical details.

Further, I'm sure I can think of some more instructions and clarification that would improve AI effectiveness. One thing I can think of off the bat is adding motivation context, e.g. this is a math website, we have lessons in video/written format, want to keep workflow simple, etc.

vincerubinetti:
Closes #591 

The `agents.md`/`claude.md` file has enough overlap with the `readme.md` that I feel we can just combine them. The `readme.md` could stand to be less verbose anyway, so I've made it more succinct and reorganized it all, trying to capture the actual meat of what was in `readme.md` and `claude.md`.

Haven't tested this with an agent yet.


3b1b:
Looks good. I added a symlink for CLAUDE.md to AGENTS.md, since with Claude code, it specifically loads that file into context with each command. It would no doubt see agents as well, but it's instructions may be easier to lose sight of in the broader context of the full repo.

PR #597, opened 2026-04-27
Fix test failures

vincerubinetti:
Fixes misc test failures from recent commits.

If it isn't enabled already, I'd highly recommend disallowing merge with failing tests (can still override as an admin, but it gives you a big red button warning). I think we should get used to not merging until all red ❌s are resolved one way or another. Not saying all the tests are 100% correct or necessary, so sometimes the solution will be reworking or removing the test.

But please do check with me (or someone) before removing something, as its purpose might not be readily apparent. E.g., the order of Tailwind classes in `className` attrs might seem persnickety, but I've found that it prevents a lot of potentially mission critical issues by making it obvious if you have duplicate styles or styles that will overwrite other ones.

The Axe tests in particular I could see being very annoying and difficult to fix unless you have experience, but those are important. And as a tip: the Playwright report for failing tests has a very handy "copy prompt" button, and I've found Claude very effective at explaining and fixing issues from it:

Screenshot 2026-04-27 at 12 22 47 PM

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,1 @@
-routes.json
+*.json
--- a/app/pages/talent/doppel/index.mdx
+++ b/app/pages/talent/doppel/index.mdx
@@ -18,7 +18,7 @@ import rsaconf from "./rsaconf.jpg";
 import threatWorkflow from "./threat-workflow.webp";
 import wordmark from "./wordmark.svg";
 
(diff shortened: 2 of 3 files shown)


3b1b:
Sorry about that, I'll keep a closer eye on that for all future merges.

Issue #592, opened 2026-04-22
Make e2e tests less flaky


vincerubinetti:
I need to figure out a way to make e2e tests more reliable. They'll fail locally and on CI at random, even after several retries. Might be fixed by tweaking how I'm waiting for each page to be "fully" loaded, or maybe by waiting for the whole Playwright test process to be fully warmed up.

PR #598, opened 2026-04-27
Fix search console warnings

3b1b:
I got two warnings from Google Search Console, one about some pages not indexing properly, one about video structure data issues.

The suggested fix for the former was to better acknowledge the "www" variant as the canonical URL. The latter probably doesn't matter immensely, but I suppose it makes it easier for video metadata to be read by crawlers.

### DIFF
--- a/app/components/Meta.tsx
+++ b/app/components/Meta.tsx
@@ -1,4 +1,5 @@
 import type { Thing, WithContext } from "schema-dts";
+import { useLocation } from "react-router";
 import { JsonLd } from "react-schemaorg";
 import { truncate } from "lodash-es";
 import site from "~/data/site.json";
@@ -18,6 +19,9 @@ export default function Meta<Type extends Thing>({
   description,
   jsonLd,
 }: Props<Type>) {
+  const { pathname } = useLocation();
+  const canonicalUrl = new URL(pathname, site.url).href;
+
   const combinedTitle = [title, site.title]
     .flat()
     .filter(Boolean)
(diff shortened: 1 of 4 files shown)


app/components/Meta.tsx
```diff
@@ -30,12 +34,13 @@ export default function Meta<Type extends Thing>({
     <>
       <title>{combinedTitle}</title>
 
+      <link rel="canonical" href={canonicalUrl} />
       <meta name="title" content={combinedTitle} />
       <meta name="description" content={combinedDescription} />
       <link rel="icon" type="image/svg" href="/icon.svg" />
 
       <meta property="og:type" content="website" />
-      <meta property="og:url" content={site.url} />
```
vincerubinetti, line 38: Maybe let's delete `url` in `site.json` then, and then I suppose the source of truth would be the Netlify redirects.

vincerubinetti:
Looks great! I just tried this today, and the lesson description meta data seems to work like a charm:

Screenshot 2026-04-27 at 1 33 58 PM

I think due to some bugs or oversights, this didn't work before the refresh. (I think it would've just said the default site title and description).

And I'd imagine with adding the Video object, you might get a nice video thumbnail like the first result there,

vincerubinetti:
Also FYI, if you see an e2e test failure with `Error: page.evaluate: Execution context was destroyed, most likely because of a navigation`, that's something I still need to fix, tracked in #592. Simply re-running the failure may resolve the error, but otherwise you can ignore for now.

app/components/Meta.tsx
3b1b: In that case, would you be alright with "https://www.3blue1brown.com" being hardcoded directly in Meta.tsx, since we still need the origin string somewhere to build the canonical href (since useLocation() only gives the pathname)?

In some ways, that's more honest about what it is (a hardcoded canonical origin).

vincerubinetti: Oh sorry I misread this line:

> const canonicalUrl = new URL(pathname, site.url).href;

I thought this was taking this directly from `window.location` (which would be redirected in the client). But that doesn't make sense anyway because we're pre-rendering and `window` wouldn't exist. Disregard my comment.

vincerubinetti (approved):

PR #643, opened 2026-06-15
Print script

vincerubinetti:
#556

- use `usePrinting` hook in addition to `print:hidden` where appropriate. mostly inconsequential, but might make layout bugs slightly less likely by removing the element completely instead of hiding with css. and makes page rendering slightly faster.
- change display `flex` to `block` for certain elements when printing, because Chrome gets thrown off when using `flex` and page breaks when printing. Firefox and Safari handle it fine, so if some day we want to keep the precise-ness of flex spacing in the prints and are okay with having to locally install those browsers, we can get rid of the `print:block` overrides.
- extract `waitForMath` to test util. add `waitForImages` test util. technically it's not really testing anything, but it's way more convenient to implement it as a test: can use existing test utils, playwright will handle concurrency and browser setup/teardown, etc.
- add new "test" that prints pages to pdfs using the playwright test infrastructure
- add new `bun run print` command to print lesson pages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ node_modules/
 /blob-report/
 /playwright/.cache/
 /playwright/.auth/
+
+print
--- a/app/components/Footer.tsx
+++ b/app/components/Footer.tsx
@@ -1,6 +1,9 @@
 import DarkMode from "~/components/DarkMode";
+import { usePrinting } from "~/util/hooks";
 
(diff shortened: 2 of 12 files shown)


vincerubinetti:
@SuperCowProducts Have you had a chance to look at this? I'm thinking of just merging this soon and fixing any per-lesson print layout issues later as they come up.

PR #648, opened 2026-06-27
Update redirects

3b1b:
Currently redirects only fire after a request falls through to the 404 page. This moves most redirects back to the netlify.toml page, which returns 301s, run before the /* → /404 fallback

@vincerubinetti, I wanted to get your read here, since I can tell you moved things deliberately. Am I right, it's because of the regex resolutions, combined with a desire for a single source of truth?

This was based on a conversation with Claude, which listed these downsides to the present approach:

  - Not a real redirect. Search engines and crawlers see a 404 status, not a 301/302. Link equity isn't passed, and these URLs may get deindexed. For a high-traffic site where people share links like /janestreet, that matters.
  - Requires JS + a flash. A user with JS disabled, or on a slow connection, lands on the "Not Found" page (or sees it flash) before the redirect happens. Bots without JS just see the 404.
  - Coupling. Redirect logic lives inside the 404 component, which is a bit surprising as a place to look.

### DIFF
--- a/app/pages/NotFound.tsx
+++ b/app/pages/NotFound.tsx
@@ -1,5 +1,3 @@
-import type { Route } from "./+types/NotFound";
-import { redirect } from "react-router";
 import Footer from "~/components/Footer";
 import Header from "~/components/Header";
 import { H1 } from "~/components/Heading";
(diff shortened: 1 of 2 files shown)


vincerubinetti (approved):
This is a good change. I think I started out doing it via JS so I could test it locally and to support more complex replacements, and I just forgot to revisit it.  Also I didn't realize that the Netlify redirects supported splats until very recently. Wouldn't say the 404 page is an "surprising" place to look for client-side redirects, but the 301s will be better for SEO.

Also I believe even the `/topics` redirect can be done with a splat, so we can just get rid of the client loader all together.

PR #667, opened 2026-07-30
Add tag support for talent partners

vincerubinetti:
- add `tags` as field on partner mdx files
- remove badge css util, simple enough to just do in place
- show tags in main link buttons on partners page
- add example tags to Jane Street, add example of showing them on partner page itself

### DIFF
--- a/app/components/Card.tsx
+++ b/app/components/Card.tsx
@@ -29,7 +29,7 @@ export default function Card({
     <Link
       arrow={false}
       className={clsx(
-        "group relative isolate grid rounded-md text-black no-underline outline-none hocus:scale-101",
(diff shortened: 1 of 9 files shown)


vincerubinetti:
If the general look and implementation of the tags looks good to you, let's merge now (I'll remove the Jane Street test tags) so we can get the dependent #668 and #669 (in the future I'll be more careful to not chain PRs together).

3b1b:
This works; feel free to merge.

Fwiw, at some point in the coming weeks I was thinking about taking some time to try alternate designs of the page to better feature salient information about each company (e.g., a preview of a message from me, the geography, whether they have internships, etc.)

PR #668, opened 2026-07-30
Feature interactives more

vincerubinetti:
I forked this branch from #667, so diff will show changes from that PR as well until I rebase.

- change css approach to "ribbon" (full-screen-width bg color) in interactive and home page selected topic
- move interactive fullscreen button to corner of ribbon
- rename `read` to `readable` where appropriate
- only search lessons by specified keys
- add filter button next to lesson search box to filter by readable and interactive
- tweak display and text of read/interactive buttons next to lesson results
- just compute `interactive` field on lessons automatically, like `readable`
- fix regression i noticed in inscribed rect lesson interactive involving canvas component

PR #669, opened 2026-07-30
Fix lesson nav

vincerubinetti:
Closes #661 

I forked this branch from https://github.com/3b1b/3Blue1Brown.com/pull/668, so diff will show changes from that PR as well until I rebase.

- don't show prev/next lesson buttons when lesson doesn't belong to topic
- remove "miscellaneous" topic from topics.json, derive automatically from all lessons that aren't listed in that file

### DIFF
--- a/app/data/topics.json
+++ b/app/data/topics.json
@@ -283,26 +283,5 @@
       "discrete-fourier-transform",
       "diffusion-equation"
     ]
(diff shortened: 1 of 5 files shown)


3b1b:
This feels like the right solution, thanks!

PR #673, opened 2026-08-05
Talent page redesign

3b1b:
Proposed redesign to hoist more talent page content onto the top-level page.


app/components/Tabs.tsx
```diff
@@ -17,7 +17,7 @@ type Props = {
 export default function Tabs({ tabs, children, className }: Props) {
   return (
     <_Tabs.Root className={clsx("flex flex-col items-center gap-4", className)}>
-      <_Tabs.List className="relative flex flex-wrap items-center justify-center gap-4">
+      <_Tabs.List className="relative flex items-center justify-center gap-4">
```
vincerubinetti: I'm guessing this was to avoid wrapping on small screens:

Image

But personally I find this unpleasant:

Image

What about just `flex-wrap gap-x-4 gap-y-2` to recover a bit of vertical space:

Image

I tend to lean and encourage letting things breathe and not being too precious about whitespace.

But here, either way is fine, it's just preference.

app/components/Vimeo.tsx
```diff
@@ -57,7 +57,14 @@ export default function Vimeo({
         aria-label="Play video"
         className={clsx(className, "group")}
       >
-        <img src={thumbnail} alt="" className="size-full object-cover" />
+        <img
+          src={thumbnail}
+          alt=""
+          // pages can stack many of these, so only fetch the ones on screen
+          loading="lazy"
+          decoding="async"
```
vincerubinetti, line 64: Probably a micro-optimization, but I suppose YT/Vimeo could serve a huge jpg that could take an older device longer to decode.

app/components/Tabs.tsx
vincerubinetti: Actually we probably do want wrapping in case you have a many-tab use-case in the future.

`overflow-x` with no wrap is another option.

app/pages/talent/PartnersHeader.tsx
```diff
@@ -1,16 +1,42 @@
...
+        </H1>
+        <div className="text-center font-sans text-2xl text-balance">
+          A virtual career fair
+        </div>
+      </hgroup>
+
+      <Button
+        size="sm"
+        onClick={onToggleIntro}
+        aria-expanded={showIntro}
```
vincerubinetti: I think with the changing of the talent slogan, it's more appropriate to hide this explainer.

And while this should be accessible with the aria attributes, it still feels weird to me. I don't have a concrete reason why, other than I've just never seen a button in a header make another section on the page appear/disappear.

I'm going to commit an alternate suggestion to see if you like it.

app/pages/talent/Talent.tsx
```diff
@@ -19,11 +23,14 @@ export default function Talent() {
       <Header
         background={<TriangleGrid className="mask-b-from-0% mask-b-to-100%" />}
       >
-        <PartnersHeader />
+        <PartnersHeader
+          showIntro={showIntro}
```
vincerubinetti: If we do keep this cross-section toggle, it might just be better to use an atom defined in the intro section. Not react-idiomatic, but OK here in this simple case, and cleaner sharing of state across the component tree instead of hoisting the state up to the common parent.

app/pages/talent/Intro.tsx
```diff
@@ -1,17 +1,25 @@
+import clsx from "clsx";
 import PiCreature from "~/components/PiCreature";
 
+// id of this section, so the header's toggle can reference it
+export const introId = "talent-intro";
+
+type Props = {
+  // hide the section, but keep it mounted for the header's toggle to reference
+  hidden?: boolean;
```
vincerubinetti: Not wrong per se, but similar how I commented about the `wordmarkScale` feeling a code smell by being too fine-grained and one-off, maybe just forward `className` like many other components do instead of a toggle prop for one specific style.

app/pages/talent/Partners.tsx
```diff
@@ -44,30 +38,13 @@ export default function Partners() {
...
-              <div className="mt-4 font-sans text-2xl font-medium">{name}</div>
-              <div className="text-center text-balance text-gray">
-                {tagline}
-              </div>
-              <Tags tags={tags} className="mt-2 text-sm" />
-            </Card>
-          );
-        })}
+      {/* dividers match the site's hr, which is also bg-current/10 */}
+      <div className="flex w-full flex-col divide-y divide-current/10">
```
vincerubinetti: Can be simplified by just using section's natural flex-col and gap behavior.

app/pages/talent/PartnerTabs.tsx
```diff
@@ -0,0 +1,89 @@
+import type { ReactNode } from "react";
+import ShowPartial from "~/components/ShowPartial";
+import Tabs, { Panel } from "~/components/Tabs";
+import Vimeo from "~/components/Vimeo";
+import YouTube from "~/components/YouTube";
+import { getChallenge, getMessage, getPartner } from "./Partner";
+
+// mdx blocks stack with gaps, like the sections they'd sit in on a partner page
+const prose = "flex w-full flex-col gap-4";
```
vincerubinetti: For such a short string and simple style, I'd normally just lean towards naively copy-pasting it. I don't make an effort to extract basic style repetition elsewhere on the site. Also if I do have a `className` variable hoisted, I've kept it within the component scope elsewhere.

vincerubinetti (approved):
Overall, looks great. Made some suggested changes, feel free to revert.

app/pages/talent/Partners.tsx
```diff
@@ -44,31 +39,15 @@ export default function Partners() {
...
-              <Tags tags={tags} className="mt-2 text-sm" />
-            </Card>
-          );
-        })}
-      </div>
+      {order.map((id, index) => (
+        <Fragment key={index}>
+          <PartnerRow
+            id={id}
+            className={clsx(index % 2 === 1 && "flex-row-reverse")}
```
vincerubinetti: If you don't like the alternating order, just remove this line and the `className` prop on `PartnerRow`.

app/pages/talent/shopify/index.mdx
```diff
@@ -32,14 +35,10 @@ import wordmark from "./wordmark.svg";
...
-<YouTube id="_VqBW7h5BYg" />
+<YouTube id={frontmatter.youtube} />
 
-<Button
-  to="https://www.shopify.com/careers?utm_source=3b1b"
-  color="critical"
-  className="w-100 self-center"
->
-  Explore Open Roles
+<Button to={frontmatter.apply} color="critical" className="w-100 self-center">
```
vincerubinetti, line 40: Changed these to take directly from frontmatter.

app/pages/talent/PartnerTabs.tsx
```diff
@@ -0,0 +1,81 @@
...
+      ),
+    });
+  else if (quote)
+    tabs.push({
+      label: `About ${name}`,
+      content: <p>{quote}</p>,
+    });
+
+  // technical challenge or puzzle
+  if (Challenge)
```
vincerubinetti, line 65: More of an editorial comment: are you sure you want this much detail on this page? It feels like mixing high-level and low-level.

The challenges in particular get quite long, and feel to me like they should remain on the dedicated pages:

Image

Image

A different approach for your consideration:

Image

app/pages/talent/PartnerRow.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const wordmark = darkMode ? getWordmarkDark(id) : getWordmark(id);
+  const to = href("/talent/:id", { id });
+
+  return (
+    <div className="grid grid-cols-[minmax(0,1fr)_minmax(0,1.25fr)] items-start gap-x-12 gap-y-6 max-md:grid-cols-1">
+      {/* identity */}
+      <div className="flex flex-col items-center gap-4 text-center">
+        <Link
+          to={to}
+          arrow={false}
```
vincerubinetti: Not necessary because internal link. Link component will detect and not show arrow.

app/pages/talent/PartnerTabs.tsx
3b1b: Not a bad suggestion, and I hear you on mixing high-level vs. low-level. In fact, you're suggestion is definitely the cleaner of the two options.

As a bit more backdrop, one thought in my mind with the redesign is to eventually question whether it makes sense to have the dedicated partner subpages at all, if they have a lot of overlap with what's on that company's own website / about / careers page anyway. In that case, the "learn more" would point to a partner's own website.

The philosophy of the resign is to better surface what is more personal and specific to 3b1b. That is, draw a distinction between what content/information one would only find on this page, and put that more front and center. At least as an experiment, I want to see what it can feel like for all that to be _entirely_ explorable on the top-level page.

My own bias here is that I don't think it's that problematic to have a slightly larger bit of text unfold from one of the tabs, as long as it's something the user would want to see. I think I also have a bias that a click to a separate page, while super-low-friction in principle, actually does act as a barrier for someone just poking around an exploring. That is, realistically fewer people will actually see the challenges if they show up as blue links, as opposed to as tabs.

app/components/ShowPartial.tsx
```diff
@@ -17,11 +19,15 @@ type Props = {
...
+export default function ShowPartial({
+  defaultOpen = false,
+  className,
+  children,
+}: Props) {
   const ref = useRef<HTMLDivElement>(null);
 
   // state
-  const [open, setOpen] = useState(false);
+  const [open, setOpen] = useState(defaultOpen);
```
vincerubinetti, line 30: Oh yes, I forgot to fix the bug with `useAutoHeight` where it would temporarily be full height before collapsing upon focusing the tab.

If that was the only motivation behind this change, I can just fix it. Having a defaultopen prop is good though.

3b1b: I'm not sure I see that bug. I believe this was part of the change where if someone tabs over to a bit of text that may need to be expanded, it starts its life expanded. That is, if someone has opted into it, you can assume they're interested in reading it rather than asking to tap again to expand.

app/pages/talent/PartnerTabs.tsx
vincerubinetti: Makes sense re: wanting to surface more info sooner. A single click definitely does cut down follow through dramatically. Note that the tabs still require an extra click, though it still allows the other partners to stay in view.

I'm loathed to say it because I hate popups as a user, but you might want to also try a `<Dialog>` for the message/challenge here instead of `<Tabs>`, so the large content can fill the screen more without affecting the layout of everything else (and avoid the ugly e.g. `md:max-h-88`).

app/components/ShowPartial.tsx
vincerubinetti: auto-height-bug


I should fix this anyway.

PR #678, opened 2026-08-08
Form submission tweaks, email pattern, Windows dev checks

johhhnnnnyyyyy:
- add error condition as check for disabling feedback submit button
- style blank `<Select>` option specially
- add regex pattern check to email box
- lazy-import vimeo component to avoid [rare upstream bug](https://github.com/vimeo/player.js/issues/1153), and because we only use it on a few pages anyway
- for consistency with feedback form, keep contact form submit button visible and just disable
- check that we can run site locally on windows under WSL


johhhnnnnyyyyy:
Hey! Following up on the email validation task, I've implemented the fix and wanted to share what I did and flag one thing before opening a formal PR.

## What I changed

### `app/components/TextBox.tsx`

- Added an `EMAIL_PATTERN` regex constant at the top of the file defined once,  applied automatically.
- When `type="email"` is passed to `TextBox`, the `pattern` and `title` attributes are automatically set on the underlying `<input>`  no changes needed at any call-site.
- The browser's **native validation tooltip**  handles all user feedback no custom UI, no extra state.

### `app/pages/about/Contact.tsx`

- The Send button was conditionally hidden (`{!status && <Button>}`) which meant it disappeared after a failed submission, forcing users to reload before they could try again.
- Changed it to always render, using `disabled` to control interactivity instead:
  - **Disabled** during `"loading"` — prevents double-submit.
  - **Disabled** after `"success"` — form is done.
  - **Enabled** after `"error"` — user can fix their input and retry without reloading.

## Testing without a live Netlify deployment

Submitting with all correct inputs will still show the error alert locally as the form POSTs to Netlify's processing endpoint which doesn't exist in local dev. The actual end-to-end submission can only be verified on a Netlify deploy preview.

I currently used relative paths (e.g., ../../components/TextBox) in my changes for running the website. I plan to push a quick update to switch these to ~/ to keep it consistent with the rest of the codebase, unless anyone prefers otherwise.

Thanks so much for the opportunity to contribute to 3Blue1Brown! Please let me know if you need any adjustments or formatting changes.

app/components/MathJax.css
```diff
@@ -1,4 +1,4 @@
-@reference "~/styles.css";
+@reference "../styles.css";
```
vincerubinetti: Per the style guides in agents.md:

> Use `~/` prefix for all imports (except when files meant to always be colocated)

app/components/TextBox.tsx
```diff
@@ -32,6 +32,9 @@ type Multi = {
   multi: true;
 } & Omit<ComponentPropsWithRef<"textarea">, "value" | "onChange">;
 
+const EMAIL_PATTERN =
```
vincerubinetti: In my browser I'm seeing that this is an invalid regex:

Image

I think this regex is overly complex as well. Simplicity is highly prioritized in this repo. Especially since I think the [value of validating email here is low](https://stackoverflow.com/a/202528/2180570).

The built-in `type="email"` validation for chrome seems to be something simple like `something@something`. I wouldn't want to be much more restrictive than that. I was imagining this PR might just add a check for a period after the `@` or something like that.

This site is viewed by people all over the world, and though there is an RFC, the traditional wisdom is that in practice email addresses can vary more widely and if you strictly enforce the standard you'll block a lot of legitimate addresses.

app/components/TextBox.tsx
```diff
@@ -49,6 +52,8 @@ export default function TextBox({
   const mergedRef = useMergedRefs(inputRef, passedRef);
   const sideRef = useRef<HTMLDivElement>(null);
 
+  const isEmail = !multi && (props as Single).type === "email";
```
vincerubinetti: Can avoid casting by doing `!!multi && "type" in props && props.type === "email";`

Style guide:

> Avoid unsafe TypeScript: `any` (use `unknown` instead), casting (`as SomeType`), non-null assertions (`value!`)

app/components/TextBox.tsx
```diff
@@ -91,8 +95,13 @@ export default function TextBox({
       className={clsx(inputClass, "h-12")}
       style={{ paddingRight: sidePadding ? sidePadding : "" }}
       value={value}
+      pattern={isEmail ? EMAIL_PATTERN : undefined}
+      title={
```
vincerubinetti: This wouldn't be a very accessible error message, we'd want to use something like `setCustomValidity` or `role="alert"`.

app/pages/about/Contact.tsx
```diff
@@ -129,16 +129,15 @@ export default function Contact() {
...
-        <Button
-          type="submit"
-          color="critical"
-          className="col-span-full w-50 justify-self-center"
-        >
-          <PaperPlaneIcon />
-          Send
-        </Button>
-      )}
+      <Button
```
vincerubinetti, line 132: > which meant it disappeared after a failed submission, forcing users to reload before they could try again.

We want to slow the user down here, as hinted at by the error message "please try again later". In practice, errors would occur here either from a bug in the website code, a server outage, or rate limiting, none of which would benefit from retrying immediately. We also definitely don't want to let people (intentionally or unintentionally) spam messages.

app/pages/about/Contact.tsx
```diff
@@ -129,16 +129,15 @@ export default function Contact() {
...
+        type="submit"
+        color="critical"
+        className="col-span-full w-50 justify-self-center"
+        disabled={status === "loading" || status === "success"}
```
vincerubinetti: This website makes use of `aria-disabled` instead of `disabled` for accessibility reasons.

vite.config.ts
```diff
@@ -30,7 +31,13 @@ export default defineConfig(() => ({
   resolve: {
     tsconfigPaths: true,
     alias: {
-      "~/": new URL("./app/", import.meta.url).pathname,
+      // fileURLToPath avoids the leading slash that .pathname gives on Windows
+      // (which doubles the drive letter: C:\C:\...). replaceAll normalizes
+      // backslashes to forward slashes so Vite's SSR module runner can resolve them.
+      "~/": fileURLToPath(new URL("./app/", import.meta.url)).replaceAll(
```
vincerubinetti: I will use this PR as an opportunity to test the website dev mode fully on my Windows machine, since we have thus far basically only run it on Macs. I'll do this later today or tomorrow.

vite.config.ts
```diff
@@ -104,7 +111,10 @@ const mdxPlugin = mdx({
     editMDX,
   ],
   // https://mdxjs.com/packages/mdx
-  providerImportSource: "~/components/Markdownify",
+  // MDX resolves providerImportSource before Vite aliases, so use an absolute path
```
vincerubinetti: Though this change seems to be necessary after your Windows fix above, I'm not sure this comment is exactly accurate as to the reason. I'll need to look into this.

app/pages/about/Contact.tsx
```diff
@@ -11,7 +11,7 @@ import TextBox from "~/components/TextBox";
 const formName = "contact-general";
 
 const reasonOptions = [
-  { value: "", label: "" },
+  { value: "", label: "Select a Reason" },
```
vincerubinetti: Not sure this is necessary or wise. At a glance, this makes it looks like an option is already selected. I think seeing "Reason", the \* required asterisk, and a blank box makes it clear that something still needs to be done. If we were to do this, I'd want to style the label as `text-gray` when `value` === `""` in the Select component, to indicate its blankness.

vincerubinetti (commented):
If you have the option, please select "Allow edits by maintainers" so I can push changes.

app/pages/about/Contact.tsx
johhhnnnnyyyyy: Yeah, its optional just added it for better UX

vite.config.ts
johhhnnnnyyyyy: Sure, thanks

app/pages/about/Contact.tsx
vincerubinetti: For consistency with the `<Feedback>` form, we can let the button always be visible but disabled when needed. But in both cases it should be disabled on loading, error, or success. I just noticed that "error" is missing from the feedback, which is an oversight.

johhhnnnnyyyyy: Oh, thanks for the information.

johhhnnnnyyyyy:
> If you have the option, please select "Allow edits by maintainers" so I can push changes.

"Allow edits by maintainers" option is currently checked on my end. GitHub might be experiencing a UI lag, or it could be an issue with permissions on the fork.

vite.config.ts
vincerubinetti: Looking into this now. Can you provide me with more details, like: which version of bun and node you have installed, are you running `bun run dev` through WSL or not, what error messages did you get that these `vite.config.js` changes are attempting to resolve, etc.

vite.config.ts
vincerubinetti: Okay, I asked Claude to investigate the call train here, and it's as I expected: this option just literally adds a `import ... from "~/components/Markdownify"` to the top of MDX files, which Vite comes in later and resolves correctly with the alias.

Although I said this looked to be necessary due to your other change in `vite.config.ts`, I can no longer replicate that... this line works fine on both Mac and Windows with the `~/` alias. Perhaps it was made necessary due to another change that I have since reverted.

vite.config.ts
vincerubinetti: Okay, I tested out the site on Windows, and have everything working now.

I actually found that the original code here works, but I am using WSL, and perhaps you aren't. I think we will require that Windows users need WSL to run the site locally; full native Windows compatibility is just too onerous to be worth it (sorry).

Still, I've kept the `filePathToUrl` change since it's more robust and "correct". But unless you were running into a reproducible error on WSL, the `replaceAll` shouldn't be necessary. The filePath util func should handle all the normalization and compatibility itself.

Note that I also ran into an unrelated issue with the Vimeo component due to an upstream bug when only Bun is installed and Node is not:
https://github.com/vimeo/player.js/issues/1153

vincerubinetti (approved):

PR #679, opened 2026-08-08
Talent page cleanup

vincerubinetti:
- change tabs component api shape: tab title defined right on panel content instead of as separate prop. this makes conditional rendering cleaner.
- change all usage points of tabs component
- get rid of ShowPartial around "message from grant"?
- rename some components
- merge PartnerTabs just back directly into PartnerTile now that it's only used there
- simplify partner tile css, polish alignment

### DIFF
--- a/app/pages/talent/Gallery.tsx
+++ b/app/pages/talent/Gallery.tsx
@@ -1,7 +1,7 @@
 import { Fragment, useEffect, useState } from "react";
 import { H2 } from "~/components/Heading";
 import { seededShuffle } from "~/util/math";
-import PartnerRow from "./PartnerRow";
+import PartnerTile from "./PartnerTile";
 
 const partners: [string, ...string[]] = [
   "janestreet",
@@ -16,7 +16,7 @@ const partners: [string, ...string[]] = [
 ];
 
 // gallery of partners
-export default function Partners() {
+export default function Gallery() {
   const [order, setOrder] = useState(partners);
 
   useEffect(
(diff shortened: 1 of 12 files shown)


app/pages/talent/PartnerTabs.tsx
```diff
@@ -29,56 +28,43 @@ export default function PartnerTabs({ id, className }: Props) {
...
-      ),
-    });
-
-  // message from grant
-  if (Message)
-    tabs.push({
-      label: "Message from Grant",
-      content: (
-        // clicking over to this tab is already intent to read it
-        <ShowPartial defaultOpen={hasVideo}>
```
vincerubinetti, line 53: I figured this wrapping show partial is not necessary if the content is already hidden by the tab?

app/pages/talent/PartnerTile.tsx
```diff
@@ -3,16 +3,24 @@ import { ArrowUpRightIcon } from "@phosphor-icons/react";
...
+  getWordmarkDark,
+} from "./Partner";
 
 type Props = {
   // partner id
   id: string;
 };
 
 // single partner in gallery
-export default function PartnerRow({ id }: Props) {
```
vincerubinetti, line 15: Maybe persnickety, but I try not to name things based on layout specifics, more on its purpose.

app/pages/talent/Gallery.tsx
```diff
@@ -16,7 +16,7 @@ const partners: [string, ...string[]] = [
 ];
 
 // gallery of partners
-export default function Partners() {
```
vincerubinetti, line 19: Found myself getting confused with too many components starting with "Partner", but feel free to revert if this is more opaque to you.

3b1b:
Thanks for the cleanup; this is better.

PR #686, opened 2026-08-12
Add a specified image to show up in social-previews of talent page

3b1b:
I noticed in a link to the [talent page](https://www.3blue1brown.com/talent/) from a Google doc that the preview image that came up was just the thumbnail of one of the videos. This change is intended to allow that preview image (or those on other pages too) to be more deliberate.


vincerubinetti (approved):
LGTM

If you want to make the image consistent with the site, you can use this:

www 3blue1brown com_talent_ (5)


For this kind of stuff, I often open the dev tools, delete the unneeded elements, and use the built-in screenshot feature (most browsers have one):

Screenshot 2026-08-12 at 2 21 51 PM

(set header to fixed size and vertical centered content)

Image

(screenshot node command)

```
h = $("header");
h.children[1].remove();
h.style.width = "800px";
h.style.height = "400px";
h.style.justifyContent = "center";
h = $("hgroup");
h.style.margin = 0;
h = $("a:has(svg)");
h.remove();
```

vincerubinetti:
I see your intermediate commits made the links absolute. Google says this is necessary too, though I've never observed a tool (e.g. posting on facebook, slack, imessage, etc.) have problems with a relative thumbnail image. But maybe it only happens with actual search engine crawlers, though I don't know what the impact of that would be (e.g. google doesn't show your thumbnail in the results).

If you did want to go back to absolute, note that we're deploying prod to Netlify too, and I'd imagine they also set an env var that has the prod URL. So we could just get rid of `site.url`.

3b1b:
I was just using the intermediate commit to test the preview link here, but I wasn't sure if it was worth the added baggage on this PR. I didn't feel in a position to judge whether the given solution was the right one.

If something like that jibes with your instincts on the right way to do it, I'll re-introduce it on a separate PR.

vincerubinetti:
If that was needed to make it work on the PR preview, but it's working now on prod, it seems like a lot of extra complexity to add for something that shouldn't change often, should reliably work (testing not really needed), and is so limited in scope.

3b1b:
Agreed, that was my thought.

PR #694, opened 2026-08-26
Update js page

3b1b:
Add whiteboarding session video to JS page


app/pages/talent/Partner.tsx
```diff
@@ -22,6 +22,12 @@ type RawPartnerFrontmatter = {
   youtube?: string;
   vimeo?: string;
   vimeoHash?: string;
+  // extra video, e.g. a whiteboard session
+  extraTitle?: string;
```
vincerubinetti: Might want to consider restructuring these less flat-ly, at this point, e.g.:

```
interview: {
  title: string;
  video: string;
  type: "youtube" | "vimeo"
},
whiteboarding: {
  ...
}
````

3b1b: Good point

PR #697, opened 2026-08-28
Added Oklo, Luminal, and MatX pages

SmashleyHP:
Added Oklo page (from provided materials) and Luminal and MatX pages (both from placeholder materials while we wait for them to be provided)


SmashleyHP:
Note that I asked Oklo for a revised banner image because of the cutoff text!

vincerubinetti:
FYI the test is just complaining that the Luminal office picture is too large in filesize. Should probably be converted to jpg and downsized a bit.

Currently to view the logs of the end-to-end tests you have to download the artifact and open the HTML file in a browser:

Screenshot 2026-08-31 at 10 45 31 AM

I'll look into making that easier though. I think there's some plugins that can automatically post a link to the logs as a comment on the PR.

PR #698, opened 2026-08-31
Comment on pr with details about e2e test failures

vincerubinetti:
This should make it easier to see and fix end-to-end test failures by posting the details in a comment on the PR. ~I added an intentional error on the home page to test this change, which I should remove before merge.~

It's deleted now because I removed the fake error, but here's a taste of the comment that now gets generated when there's an e2e failure:

Screenshot 2026-08-31 at 2 44 58 PM


3b1b:
Thanks! This should be helpful.

PR #700, opened 2026-09-01
Fix squished Figure images in Safari on Talent partner pages

SmashleyHP:
Captionless Figure images render a bare  as a direct flex item of the page section, which hits a Safari bug where percentage width + auto height loses the intrinsic aspect ratio once the section's width is clamped. Wrapping in a block-level div keeps the img out of the flex layout and avoids it.

### DIFF
--- a/app/components/Figure.tsx
+++ b/app/components/Figure.tsx
@@ -30,19 +30,17 @@ export default function Figure({
   children,
 }: Props) {
   // image to render
-  //
-  // when there's no caption, Image renders a bare <img>, which (as a direct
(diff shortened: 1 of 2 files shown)


vincerubinetti:
Will take a look at this soon. Need to be very careful when editing `/components`, since they're used all over the site. Especially with `<Figure>` which is in all of the written lessons. There might be a better way to solve this without a wrapping div. In general I try to avoid adding wrappers wherever possible, as they can make layout more unpredictable and difficult.

PR #701, opened 2026-09-01
Run e2e tests in webkit to catch Safari-only layout bugs

3b1b:
Testing only in Chromium missed a WebKit layout bug reported by a reader on Safari: an image with a percentage width that's a direct flex item loses its aspect ratio, rendering up to 2x too tall once the viewport is wide enough. It affected several talent partner pages and lessons, and was invisible in Chromium and CI.

- enable the webkit project, already stubbed out in the config
- install webkit alongside chromium so CI picks it up
- add an image spec asserting images render at their natural aspect   ratio, at a viewport wide enough to trigger the bug

Unlike the other route specs, the image spec doesn't skip to chromium, since layout is exactly the thing that differs between engines. It ignores images that legitimately don't match their natural ratio: unloaded, cropped/letterboxed, or padded.

Depends on #700, which fixes the bug this catches — CI here will be red until that merges.


3b1b:
@vincerubinetti, this was Claude's suggestion on how to catch bugs like the one https://github.com/3b1b/3Blue1Brown.com/pull/700 aims to address. I'd don't feel well suited to assess style/quality for tests here; if it's slop or bloat, feel free to just say so, but I wanted to get your eyes on it.

vincerubinetti:
I reworked the test to match repo style and also include videos. 

On one hand, images/videos are critical on this site. On the other, it's many extra LOC, for something very specific and more fine-grain than other tests. Also, an image being stretched isn't the end of the world as long as it is visible and not stretched so much as to be unreadable. If it could somehow catch other undesirable image rendering, it'd be great, but that's pretty ambiguous and subjective.

What might catch more errors here is a process change, e.g. PR checklist to make sure we check our changes in Chrome/Firefox/Safari.

Side note: Claude doesn't seem to respect the instruction I put in `agents.md` about that. So far, I haven't really documented high level design decisions in the code comments. I instead do that in PR text, and leave code comments as minimal reminders, enough for a knowledgeable frontend person to reconstruct the meaning. This is subjective though -- we could pivot to a different approach -- but that's what I've been doing so far.

PR #703, opened 2026-09-02
simpler fix for image stretch

vincerubinetti:
Supercedes #700

I believe this still fixes the issue, which I was primarily seeing here:

Screenshot 2026-09-02 at 11 20 09 AM

Spot checking other instances of `<Figure />`

I think this is a good example of how LLMs try to fix by adding instead of removing. The `w-full` on the image element may have been a vestigial, unnecessary style. I was probably trying to make sure even small images (e.g. 100px wide) would still stretch to fill the width of the container. But since the container (whether it be a `<section>` or a tab `<Panel>`) should always be a `flex flex-col` container, and since flex defaults to `align-items: stretch`, the container should ensure full width anyway. Even if the container had `items-center`, you could apply `self-stretch` to the image to ensure it gets full width.

This may be a small simplification in the grand scheme of things, but I view my role in this as fighting against slow accumulation of entropy in the code, to avoid ending up with workaround's piled on top of workarounds.

As for why Safari does this, that's still a mystery to me. But I've often found that over-specifying things (e.g. adding `w-full` when the parent would already ensure that) can confuse the browser more, especially Safari.

### DIFF
--- a/app/components/Figure.tsx
+++ b/app/components/Figure.tsx
@@ -1,6 +1,5 @@
 import type { ReactNode } from "react";
 import { ImageIcon, VideoIcon } from "@phosphor-icons/react";
-import clsx from "clsx";
 import Image from "~/components/Image";
 import Tabs, { Panel } from "~/components/Tabs";
 import { usePrinting } from "~/util/hooks";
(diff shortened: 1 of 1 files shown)


3b1b:
It looks like the reason the agent didn't do this was that it was testing by printing pages as PDFs, and this alters the results there. That's something I don't think about much and am not sure people even do it, but if we care, here's the proposal in line with what you have here.

```
const imageElement = (
  <Image image={image ?? ""} className={clsx("print:w-full", className)}>
    {children}
  </Image>
);
```

vincerubinetti:
Ah okay that's interesting. I guess that's one way for the agent to see the results. I've also seen it able to run the dev server locally itself and take a screenshot to inspect, which is better. Not sure how to force Copilot/Claude to do it that way though.

FYI in printing (at least with the `print` script I made) we're using Chromium. Also yeah in printed form, if an image is intrinsically too small to fill the page, that shouldn't matter too much. The page layout would switch from `flex` to `block` (to accommodate Chrome's print behavior not liking flex, as discussed in #643), and the image would just be left aligned (or center if there's a `text-center` on an ancestor somewhere, I forget).

So this PR seems to be in order.

